PR #35, opened 2022-07-15
20 | Moving from setup.py to setup.cfg and pyproject.toml

### LINKED ISSUE #20: Move from setup.py to setup.cfg
The latest iteration of Python's [setuptools](https://setuptools.pypa.io/en/latest/userguide/) deprecates the use of `setup.py` except in some specific circumstances (see [note](https://setuptools.pypa.io/en/latest/userguide/quickstart.html#setup-py)).

Instead it uses a combination of `pyproject.toml` and `setup.cfg` to achieve the configuration of the package, its dependencies and usage.

The settings in `setup.py` should be migrated to `pyproject.toml` and `setup.cfg`.

ns-rse:
Resolves #20 

Newer versions of [setuptools]() have deprecated `setup.py` in all but a few instances and instead use a combination of `setup.cfg` and `pyrpoject.toml` to configure installation of packages.

This pull request migrates the configuration defined in `setup.py` to these two files.


jonbarker68 (changes requested):
Looks good except
- hydra-core>=1.1.1 - the newer hydra 1.2 has changes in the way that output directories are handled that messes up the CEC2 recipes. This could be fixed in the recipes in the next release but for now it would be more expedient to tie to hydra 1.1.1. i.e. hydra-core==1.1.1
- "matplotlib>=3.5.1" -> "matplotlib". -- seems it is best not to pin this. e.g. in colab matplotlib 3.2.2 is pre-installed and later versions do not work in the colab environment
- version 2.0.3 -  I'd like to start again with this as version 0.1.0 and then have 0.2.0, 0.3.0 etc for each challenge round. None of the previous versions were released on pypi so I don't think this will cause any confusion.

jonbarker68 (approved):
All looks good to me :)

jonbarker68:
Not sure why the 'lint and pre-commit checks' is reported as 'expected' but doesn't seem to be running...

PR #47, opened 2022-07-23
Fixed coverage config in setup.cfg

jonbarker68:
This fix so that pytest-cov reports on all source files, not just those imported in tests.


ns-rse (approved):
Thanks @jonbarker68, I'd never clocked this as being necessary.

PR #55, opened 2022-08-09
Fix mc_conv_tasnet version incompatible issue

TuZehai:
The new PyTorch version 1.12.0 does not support four order input to conv1d in the SpatialEncoder, fix it by squeezing the redundant dimention.


jonbarker68 (approved):
All OK. Thanks.

PR #66, opened 2022-08-13
Fixed numpy typehint in compressor.py

jonbarker68:
The typehint " Tuple[Any, np.ndarray[Any, np.dtype[Any]], List[Any]]"

Was causing TypeError:

TypeError: 'numpy._DTypeMeta' object is not subscriptable

in Python 3.7.

Simplified to Tuple[Any, np.ndarray, List[Any]]


jonbarker68:
Overriding need for review as it's just one file that urgently needs to be fixed as it's breaking the colab tutorials.

PR #71, opened 2022-08-14
69 deprecation warnings need dealing with

jonbarker68:
Fixing deprecation warnings that have been exposed by the test suite.


jonbarker68:
@ns-rse this PR is just dealing with some minor deprecation warnings that were caught by the test suite. Nothing very important but some of them would have led to problems later. We might as well merge this into main before releasing v0.1.0

ns-rse (approved):
Agree that its good to get these resolved earlier. All very minor changes as you wrote.

PR #72, opened 2022-08-15
53 | import annotations from future to tests pass under Python < 3.9

ns-rse:
Using `from __future__ import annotations` so that type hints work under older versions of Python.


ns-rse:
Very strange, I've now seen the following error message under [Python 3.8](https://github.com/claritychallenge/clarity/runs/7835527554?check_suite_focus=true#step:6:80), [Python 3.9](https://github.com/claritychallenge/clarity/runs/7836587648?check_suite_focus=true) and [Python 3.10](https://github.com/claritychallenge/clarity/runs/7836706238?check_suite_focus=true) but on other runs the tests have passed without any problem its one of the other versions that fails. I don't get this error locally on a fresh 3.10 or 3.7 install with all packages/dependencies installed anew.

```
tests/regression/test_predictors.py F.                                   [100%]

=================================== FAILURES ===================================
_____________________________ test_torch_msbg_stoi _____________________________


regression test output differences for tests/regression/test_predictors.py::test_torch_msbg_stoi:

>   --- current
>   +++ tobe
>   @@ -1,2 +1,2 @@
>   -Torch MSBG STOILoss -0.46198, ESTOILoss -0.32999
>   +Torch MSBG STOILoss -0.46198, ESTOILoss -0.33000
>    
```

Its clearly a precision issue, and manually re-running the failed test it passed ok, but highlighting as it may crop up again in the future.

PR #74, opened 2022-08-15
73 msbg code requires refactoring

jonbarker68:
Some work to improve the msbg code increases the pylint score, adding further typehints and more tests.


clarity/enhancer/nalr.py
```diff
@@ -96,8 +96,8 @@ def fir2(
             ne = int(np.fix(ff[i + 1] * npt)) - 1
 
         j = np.arange(nb, ne + 1)
-        inc = 0.0 if nb == ne else (j - nb) / (ne - nb)
-        H[nb : (ne + 1)] = (inc * aa[i + 1]) + ((1 - inc) * aa[i])  # type: ignore
+        inc: Union[float, np.ndarray] = 0.0 if nb == ne else (j - nb) / (ne - nb)
```
jonbarker68, line 99: j can be a scalar or an array, so adding the "Union[float, np.ndarray]" has fixed the mypy error.

ns-rse (approved):
Great work tidying these classes up @jonbarker68

Issue #49, opened 2022-07-25
Automate versioning and releasing to PyPI via GH Actions


ns-rse:
As we move to tagged versions and releases to PyPI we need to automate the flow of versioning the package to avoid the work of updating `setup.cfg` and `clarity/__init__.py`.

The [versioneer](https://github.com/python-versioneer/python-versioneer) package appears to offer a solution to this as it automates aligning the version with GitHub tags/releases.

After the initial `0.1.0` release has been made manually the process should shift to using `versioneer` to handle versions.

ns-rse:
Closed by #78 making a Release and applying a new tag to be created triggers the `./.github/workflows/pypi.yaml` which publishes to PyPI

Releasing works with the .github/workflows/pypi.yaml and "all" you need to do is...

1) Go to the Releases page (https://github.com/claritychallenge/clarity/releases)
2) Draft a New release
3) Create a new tag using semantic versioning and select "Create new tag v#.#.# on publish"
4) Click the "Generate Release Notes" button, this adds all the titles for all Pull Requests which I've remove but includes a link to the full ChangeLog which I've left in.
5) Write your release notes.
6) Select "Set as latest release"
7) Select "Create a discussion for this releases" and select "Announcements"
8) Click on "Publish Release"

The .github/workflows/pypi.yaml then kicks into action and builds the packages then uploads them to PyPI (above notes are as much for my reference as this is only the second time I've published to PyPI in this manner albeit both within a week of each other).

A few additional related tasks are to address...

* #137
* #149

PR #75, opened 2022-08-16
36 | Version and package name for release to PyPI; including recipes

ns-rse:
This PR pulls in the quite minor changes made on branch `ns-rse/36-pypi` whilst testing releasing to PyPI to...

* include package data (`*.mat` and `*.yaml`).
* Ensure package name is `pyclarity` and version `0.1.0`.
* the modules under `recipe`.

It has been built and uploaded to PyPI and once merged to main the commit can be tagged with `v0.1.0`.

Prior to releasing I checked that the `ns-rse/36-pypi` branch that had been uploaded to TestPyPI could be installed in a clean Conda environment using...

```
❱ pip install --index-url https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple/ pyclarity
```

...and then ran the three Notebooks. A little tweaking was required to some of the paths because where I started the Jupyter server differed from the location expected in the Notebook, but all the code ran.

Having uploaded a release to the main PyPI server (see [pyclarity · PyPI](https://pypi.org/project/pyclarity/)) I have repeated this, installing into a clean Conda environment (Python 3.10.4) and running the notebooks. No problems running them (although again some tweaking to the paths). :partying_face: 

To install from PyPI...

```
pip install pyclarity
pip show pyclarity
```

Once this is merged and tagged I will start work on #49 which will automate this process so that rather than manually changing the `version` in `setup.cfg` and `__version__` in `clarity/__init__.py` these will be determined automatically from tags and commits using the [versioneer](https://github.com/python-versioneer/python-versioneer) package.

One thing to consider is whether the Notebooks should be updated to reflect installation from PyPI now the package is available?

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,7 +1,6 @@
 [metadata]
-name = clarity
+name = pyclarity
 version = 0.1.0
-
 description = Tools for the Clarity Challenge
 long_description = file: README.md,
 long_description_content_type = text/markdown
@@ -95,7 +94,7 @@ pypi =
 
 [options.package_data]
 
-* = *.mat
+* = *.mat, *.yaml
 
 [flake8]
 ignore = E203, E501, W503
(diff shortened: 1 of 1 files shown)


jonbarker68 (approved):
Brilliant. Great to now have a version released on pypi. Thanks.

PR #78, opened 2022-09-23
49 | Adding versioneer and workflow to publish to PyPi

ns-rse:
Introduces [versioneer](https://github.com/python-versioneer/python-versioneer) which automatically generates versions based on tags and commits. This in turn should allow tags to be applied and the included `~/.github/workflows/pypi.yaml` workflow will build and publish the package to PyPI automatically when on `main`.

For now I have it setup to test building and pushing to `test.pypi.org`, this will fail though because the version created by versioneer will not comply to PEP440 because it includes the commit hash and how many commits ahead of the last tag the current branch is, but I need a way of testing things first.


ns-rse:
The PyPI pipeline should now work, it will attempt to build and push to Test PyPI whenever merges are made to `master` but typically these will fail because the version produced by [versioneer](https://blog.datawrapper.de/text-in-data-visualizations/) will not be PEP440 compliant if the commit does not have a tag.

Only tagged commits will attempt to be published to the main PyPI server and the tags should be compliant with PEP440 if the current nomenclature is used.

jonbarker68 (approved):
Thanks for sorting this out. Happy to approve if it's all working. We'll need to use it to bump the version for the ICASSP2023 challenge release.

PR #89, opened 2022-10-18
Fix cpc1/e029 n_ensembles

TuZehai:
Pass the n_ensembles parameter to S2STransformerBeamSearch, so that multiple ASR models can be used (used to be just the first ASR model).


ns-rse (commented):
The changes look ok to me (syntactically correct, lints fine, passes tests), but I don't know enough about the code base to be able to comment beyond that.

jonbarker68 (approved):
Minor change to fix a bug that was in the original recipe. Looks good to me.

PR #94, opened 2022-10-21
Add validation for cfs = None

groadabike:
Add cfs=None validation in build method


tests/enhancer/test_nalr.py
```diff
@@ -0,0 +1,28 @@
...
+    "audiogram, cfs, expected",
+    [
+        (
+            np.array([45, 45, 35, 45, 60, 65, 70, 65]),
+            np.array([250, 500, 1000, 2000, 3000, 4000, 6000, 8000]),
+            5.746930975720798e-05,
+        ),
+        (
+            np.array([45, 45, 35, 45, 60, 65, 70, 65]),
+            None,
```
ns-rse, line 20: I don't know if this would make sense but the parametrisation here uses the same value for `cfs` in the first test as it is set to in the second when `cfs=None` and unsurprisingly the resulting answer is the same and the test correctly identifies that a default set of values for `cfs` are used.

Would it make sense though to have _different_ values passed in for `cfs`? I don't know if there are ever scenarios where this array may differ, either in the actual values or the length of the array (looking at it the `len(cfs)` should perhaps be the same as `len(audogram)`, but if people are able to pass in different values and these two arrays are not met then an exception will likely be raised at some point and we could have a `try: ... except <Exception>: raise` block in the code and an explicit test to ensure such an exception is captured.

clarity/enhancer/nalr.py
```diff
@@ -135,15 +135,19 @@ def hl_interp(self, hl, cfs):
         hl_interpf = scipy.interpolate.interp1d(cfs, hl)
         return hl_interpf(self.aud)
 
-    def build(self, HL, cfs):
+    def build(self, HL, cfs=None):
```
ns-rse, line 138: Rather than having this as `None` you could set the default here directly and do away with the `if()` statement.

```suggestion
    def build(self, HL: np.ndarray, cfs: np.ndarray=np.array([250, 500, 1000, 2000, 3000, 4000, 6000, 8000])):
```

ns-rse (commented):
Looks straight forward, two points that might make things a little simpler.

First is simply adding a default value for `cfs` so that if one isn't explicitly passed it is used, this negates the need to have an `if:` block to set the value.

The other comment pertains to parameterisation of tests and possible exceptions that may be raised.

groadabike:
@TuZehai, in the end, I left the `if cfs is None:` conditional to prevent errors.

I think it may be ready for merging

tests/enhancer/test_nalr.py
ns-rse: Thanks @groadabike I hadn't followed through and found that.

I don't know what the general convention is in Python but I would typically expect to see a `try: ... except <ExceptionType>: ...` to capture and raise errors rather than an `assert`.

ns-rse (approved):
Good to go :+1:

tests/enhancer/test_nalr.py
groadabike: Thank you for the comment. I didn't read this class, just add the default value for the cfs parameter.

@ns-rse. @TuZehai, in your opinion, what would be better in this case?
```
def hl_interp(self, hl: np.ndarray, cfs: np.ndarray):
    if len(hl) != len(cfs): 
        raise ValueError('Hearing losses and center frequencies don't match!')
    hl_interpf = scipy.interpolate.interp1d(cfs, hl)
    return hl_interpf(self.aud)
```
or

```
def hl_interp(self, hl: np.ndarray, cfs: np.ndarray):
    try:
        hl_interpf = scipy.interpolate.interp1d(cfs, hl)
    except ValueError as verror:
        raise ValueError("Hearing losses and center frequencies don't match!")
    return hl_interpf(self.aud)
```
Personally, I prefer the first one

PR #96, opened 2022-10-26
add py as requirement for test

groadabike:
workaround for issue https://gitlab.com/uweschmitt/pytest-regtest/-/issues/19

Signed-off-by: Gerardo Roa Dabike <gerardo.roa@gmail.com>

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -66,7 +66,6 @@ tests =
   pytest-mpl
   pytest-regtest
   yamllint
-  py
 
 docs =
   sphinx
--- a/setup.cfg
+++ b/setup.cfg
@@ -60,6 +60,7 @@ tests =
   flake8
   mypy
   pre-commit
+  py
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -65,6 +65,7 @@ tests =
   pytest-mpl
   pytest-regtest
   yamllint
+  py
```
ns-rse, line 68: Overly pedantic but the list is alphabetical I think to aid reading (well it helps me at least!), so I'd have slotted this in between `pre-commit` and `pytest`.

ns-rse (approved):
Thanks for adding this work around @groadabike :+1: 

If and when the issue with `pytest-regtest` gets resolved I'll update things.

groadabike:
As requested, I changed the order of the requirements. Little by little, getting there with the best practices

PR #102, opened 2022-11-07
Nomenclature PEP8 compliance

ns-rse:
First step in changing variable names to be more descriptive and "pythonic" (i.e. PEP8 compliant).

There are a few instances such as `for x...` where I have opted to disable the `pylint`/`flake8` check as they really don't make any difference.

I've tried to expand the docstrings too but there are still some gaps where I don't know what to write. 

Another perhaps unobvious change is that I've parameterised the values used in `mbstoi.py` such that they have defaults in the function but there is also a `parameters.yaml` which is loaded and used in the tests. This will allow future users to vary the values if desired. I'm not sure if that is something that would ever be likely to be done/needed but its good to have parameters with defaults rather than hard-coded variables.

Comments/feedback welcome, particularly on missing docstrings.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,7 +19,7 @@ repos:
 #      - id: check-yaml
 
   - repo: https://github.com/psf/black
-    rev: 22.6.0
+    rev: 22.10.0
     hooks:
       - id: black
         types: [python]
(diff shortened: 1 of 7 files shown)


TuZehai (approved):
Looks all good

ns-rse:
Thanks for the review @TuZehau, I'm going to hold back and see if anyone can improve the missing docstrings before merging.

PR #103, opened 2022-11-07
97 | Change regressoin/test_predictors.py to use pytest.approx()

### LINKED ISSUE #97: [BUG] test_torch_msbg_stoi fails due to difference of 0.001
The regression test fails when running the test `tests/regression/test_predictors.py::test_torch_msbg_stoi`
having a very small difference that may result from to randomness

```
=================================== FAILURES ===================================
_____________________________ test_torch_msbg_stoi _____________________________

regression test output differences for tests/regression/test_predictors.py::test_torch_msbg_stoi:

>   --- current
>   +++ tobe
>   @@ -1,2 +1,2 @@
>   -Torch MSBG STOILoss -0.46198, ESTOILoss -0.32999
>   +Torch MSBG STOILoss -0.46198, ESTOILoss -0.33000

```

This results in PR failing some tests.
The problem can be solved by reruning the test manually one or 2 times.

ns-rse:
Closes #97

Rather than using `pytest-regtest` these regression tests (only one of which was failing) now use [`pytest.approx()`](https://docs.pytest.org/en/latest/reference.html#pytest-approx) to test equality of returned values to their expected values. This has a relative tolerance of `1e-6` and absolute tolerance of `1e-12`.

Hopefully this will now consistently pass the Continuous Integration tests.

### DIFF
--- a/tests/regression/test_predictors.py
+++ b/tests/regression/test_predictors.py
@@ -1,12 +1,12 @@
 # Regression test
-
+import pytest
 import torch
 
 from clarity.predictor.torch_msbg import MSBGHearingModel, torchloudnorm
 from clarity.predictor.torch_stoi import NegSTOILoss
 
 
-def test_torch_msbg_stoi(regtest):
+def test_torch_msbg_stoi():
 
     torch.manual_seed(0)
     torch.set_num_threads(1)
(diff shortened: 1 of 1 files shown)


TuZehai (approved):
This could solve the problem if the difference is indeed just 1e-6, though we are not sure what causes the difference...

ns-rse:
I had to relax the (relative) tolerance to `1e-4` to get the tests to pass consistently, but you're absolutely right @TuZehai that we should really understand what is causing the difference in the first place.

The strange thing is that simply re-running the tests without changing any seed values, results in the tests passing. :shrug:

groadabike (approved):
Lokk good to me.

ns-rse:
Thanks @groadabike I'd like to see if #106 solves this though as I feel its a better solution as it solves the underlying problem rather than fudging it by rounding.

jonbarker68:
Is this still needed? Has the problem now been fixed by forcing pytorch to use the CPUs during testing?  I'm finding this a bit confusing.  For PR #118 a couple of tests failed but then passed when I reran them. But would rather not fix by simply relaxing the tolerance as that's just hiding the problem.

ns-rse:
Hi @jonbarker68 ,

I'm hoping it won't be needed, as you say relaxing tolerance is a bit of a fudge and I was hoping #109 solved this by making all modules have the ability to explicitly choose the device (`cpu` or `cuda`) and then force the tests to use them.

That sounds as though it may not have worked though as your #118 failed and needed re-running.

I'll close this PR for now though and see if I can solve this properly.

PR #104, opened 2022-11-09
91 include hasqi index

groadabike:


groadabike:
@ns-rse This PR need good tests. Can you give me some guides of what would be the best way?
- They need 2 input signals (can be random)
- The final score has some variability due to randomness

The test should be similar to haspi and haaqi.

groadabike:
> > @ns-rse This PR need good tests. Can you give me some guides of what would be the best way?
> > 
> > * They need 2 input signals (can be random)
> > * The final score has some variability due to randomness
> > 
> > The test should be similar to haspi and haaqi.
> 
> Hi Gerado. There is a regtest for HASPI within the CEC2 pipeline test function.
> 
> You could use that as a template.
> 
> https://github.com/claritychallenge/clarity/blob/main/tests/regression/test_full_CEC2_pipeline.py line 116

Thank you Will.
Btw, the current output is similar to the Matlab output.
I will try to include a good unit test

jwillbailey:
I think just a regression test is needed to pass codecov.

Jon did a guide for adding the regtests to the rep. I’ll email it over.

Will

groadabike:
I just included a simple unitest using the pytest.approx

TuZehai (approved):
Looks nice, I am fine with this if it is consistent with Matlab results.

ns-rse (approved):
Sorry for the delay in getting back to you on this.

Code looks fine and all the tests pass.

The test that has been added is a regression rather than unit test since it checks the final output rather than individual functions (i.e. units). Writing unittests is a bit of an art and you would want to test different scenarios/parameters and perhaps even capture errors that could arise.

A straight translation of Matlab code is great but longer term to perhaps make the code base easier to maintain we may want to refactor into more modular, smaller function that can have unit tests written for them. That is trickier to do without first doing the translation from Matlab > Python and knowing you get the same answer though, hence why this with a regression test is essential before refactoring.

One thing I would say is that whilst its a straight translation of Matlab code the names/nomenclature are not necessarily following PEP8 conventions. This is something I've started work on on [ns-rse/98-nomenclature](https://github.com/claritychallenge/clarity/compare/main...ns-rse/98-nomenclature) where I've started renaming variables, variables/methods/classes to be PEP8 compliant.

groadabike:
> Looks nice, I am fine with this if it is consistent with Matlab results.

Yeah, the outputs are consistent

groadabike:
Running Hasqi_v2 100 times for itype 0 and 1 
https://github.com/claritychallenge/clarity/blob/9f53fd1e3a2ec01921f9e014297955c9af23c2e4/clarity/evaluator/haspi/eb.py#L105-L111

1. itype= 0
**Python -> Mean: 0.014860849104673812 - Std: 3.950009178023314e-05**
**Matlab -> Mean: 0.0148596368529389 - Std: 5.04185388879e-5**

2. itype=1
**Python -> Mean: 0.05398243543550845 - Std: 0.00013613202670926763**
**Matlab ->  Mean: 0.0539559786183452 - Std: 0.0001554092540238**

PR #101, opened 2022-11-03
Fix cuda issue in test_enhancers & test_mc_conv_tasnet

TuZehai:
In test_enhancers.py: move signal to amfir.device properly, and add cpu() before converting to numpy
In test_mc_conv_tasnet.py: downsample functions operates on cpu

PR #109, opened 2022-11-14
106 | Making device optional across modules and explicitly cpu in tests

### LINKED ISSUE #106: Errors due different devices when running Test locally.
When running `pytest --cov=clarity` locally on my laptop, some of the tests that use Pytorch fail due to different devices.
Note that I do have a GPU on my laptop.

The output error is:

> FAILED tests/regression/test_enhancers.py::test_dsp_filter - RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cpu and cuda:0! (when checking argument for argum...
FAILED tests/regression/test_mc_conv_tasnet.py::test_convtasnet - RuntimeError: Input type (torch.cuda.FloatTensor) and weight type (torch.FloatTensor) should be the same

Maybe we need to be more explicit about the devices used in the test. We could create a util.device file, similar as proposed by @ns-rse in https://github.com/claritychallenge/clarity/issues/97#issuecomment-1313549828

ns-rse:
Closes #106 

Further to #101 this PR makes the `device` on which PyTorch tensors are run configurable across three modules.

Tests are then configured to explicitly use device "cpu" in the hope that it makes them configurable.

### DIFF
--- a/clarity/enhancer/dnn/mc_conv_tasnet.py
+++ b/clarity/enhancer/dnn/mc_conv_tasnet.py
@@ -23,6 +23,7 @@ def overlap_and_add(signal, frame_step, device):
         signal: A [..., frames, frame_length] Tensor. All dimensions may be unknown,
             and rank must be at least 2.
         frame_step: An integer denoting overlap offsets. Must be less than or equal to frame_length.
+        device: Whether to use 'cpu' or 'cuda' for processing.
 
     Returns:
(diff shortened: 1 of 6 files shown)


ns-rse:
Whilst the patch coverage decreases the overall coverage remains high.

jonbarker68 (approved):
Thanks. This looks good to me.

PR #116, opened 2022-11-16
Source Separation util for Music using HDEMUCS from torchaudio

groadabike:
Tool for audio source separation using HDEMUCS.
@TuZehai @ns-rse Now that you can see the tool, is this the best place for it?


TuZehai:
Hi @groadabike,

I think it would be great if you could put this into `enhancer`, i.e., tools for enhancing the signals. Also I think it would be better if you could use something other than just `source-separation`, i.e. probably there could be other source separation model included in the future.

clarity/utils/source_separation.py
```diff
@@ -0,0 +1,101 @@
+import torch
+import torchaudio
+from torchaudio.pipelines import HDEMUCS_HIGH_MUSDB_PLUS
+from torchaudio.transforms import Fade
+
+
+def source_separation(mix, model, sample_rate, segment=10.0, overlap=0.1, device=None):
```
ns-rse, line 7: Is it possible to devise a unit-test for this function? Ideally parameterised so that the conditions of different values for `mix.ndim` and `mix.shape` are covered.

Generally there is quite a lot of steps undertaken in this function though and it would be good to break each step out into its own very small function that then has single responsibility and unit-tests can then be written for each.

For example `# Split audio into segments` could be a function on its own, as could `# Move model and audio to device` and `separated_audio = torch.zeros(...` and a few others. If all these things "hang together" then they could be grouped into a Class.

I'll have a look and feedback more tomorrow.

ns-rse (changes requested):
The convention I've learnt (which isn't necessarily "right" but works well) is that with `tests/` if there is a sub-module to nest the tests in the same structure as they are in the module and then name tests for each class function within it.

This means rather than...

`tests/test_utils/seperate_hdemucs.py`

...it should be...

`tests/utils/source_separation.py`

...and then for each function/class/method have a test with the same name prefixed by `test_` as you've done with `test_separate_hdemucs()`.

I'll look at separating out `source_operation()` into smaller functions/methods tomorrow.

clarity/utils/source_separation.py
groadabike: Good point, thank you
I just created one test to submit the PR and discuss the code and the location where to save it.
Will improve the test following your recommendations.

groadabike:
> Hi @groadabike,
> 
> I think it would be great if you could put this into `enhancer`, i.e., tools for enhancing the signals. Also I think it would be better if you could use something other than just `source-separation`, i.e. probably there could be other source separation model included in the future.
> 
> I still think the code for training and applying the DNN model (DEMUCS) should be somewhere in the `recipes`, so that ppl can modify how to train the network or how to use it. An example would be `recipes/cec1/e009_sheffield`, i.e. the network is in the `clairty` and how to apply / train the network is in the `recipes`. Still, it depends on the purpose of using this DNN model (tbh I am not quite clear why a DNN model is used for data preparation).

Ok, I can see what you mean.
I am still a bit confused about how everything is organised.

ns-rse:
The `tests/enhancer/dnn/test_hdemucs.py::test_apply_separation()` is [failing](https://github.com/claritychallenge/clarity/actions/runs/3960566055/jobs/6784885870#step:7:94) because the shape of `mix` is only two dimensions, but three are expected on line 31 of `clarity/enhancer/dnn/hdemucs.py`.

The previous step checks "_If track is mono, repeat it to match the number of channels_".

I've tried investigating this to check what the `mix.shape` is and whether this step is actually being made but (after waiting ages for Pythons Torch package and dependencies to update!) the test just hangs on my system.

However, the [.repeat()](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.repeat.html#numpy.ndarray.repeat) method is being used here to repeat/expand the `mix` array and the specific line is...

```python
    # If track is mono, repeat it to match the number of channels
    if mix.shape[1] == 1:
        mix = mix.repeat(1, 2, 1)
```

The documentation for the method `.repeat()` suggests this should perhaps be...

```python
    # If track is mono, repeat it to match the number of channels
    if mix.shape[1] == 1:
        mix = mix.repeat(2, 1)
```

...or the [`numpy.repeat()`](https://numpy.org/doc/stable/reference/generated/numpy.repeat.html#numpy.repeat) function could perhaps be used...

```python
    # If track is mono, repeat it to match the number of channels
    if mix.shape[1] == 1:
        mix = np.repeat(mix, 2, 1)
```

I've tried the former (correcting the method to only have two arguments) but the test still hangs on my system.

Are you able to check whether this solves the test failure @groadabike please?

ns-rse:
Ha, seems I was too impatient and should have waited longer.

I think there is another error here with regards to the number of channels as its the number of rows rather than columns of the Numpy array that is required, if I put some debugging `print()` in after correcting the use of `mix.repeat()` its still a mono track (if I've understood correctly).

```python
        print(f"BEFORE  mix.shape     : {mix.shape}")
        print(mix)
        if mix.shape[1] == 0:
            mix = mix.repeat(2, 1)
    
        print(f"AFTER   mix.shape     : {mix.shape}")
        print(mix)
>       batch, channels, length = mix.shape
E       ValueError: not enough values to unpack (expected 3, got 2)

clarity/enhancer/dnn/hdemucs.py:35: ValueError
-------------------------------------------------------- Captured stdout call --------------------------------------------------------
BEFORE  mix.shape     : torch.Size([1, 441000])
tensor([[0.5338, 0.7574, 0.8446,  ..., 0.8528, 0.8175, 0.0114]])
AFTER   mix.shape     : torch.Size([1, 441000])
tensor([[0.5338, 0.7574, 0.8446,  ..., 0.8528, 0.8175, 0.0114]])
```

Switching to...

```python
        print(f"BEFORE  mix.shape     : {mix.shape}")
        print(mix)
        if mix.shape[0] == 1:
            mix = mix.repeat(2, 1)
    
        print(f"AFTER   mix.shape     : {mix.shape}")
        print(mix)
>       batch, channels, length = mix.shape
E       ValueError: not enough values to unpack (expected 3, got 2)

clarity/enhancer/dnn/hdemucs.py:35: ValueError
-------------------------------------------------------- Captured stdout call --------------------------------------------------------
BEFORE  mix.shape     : torch.Size([1, 441000])
tensor([[0.0486, 0.2010, 0.7882,  ..., 0.5912, 0.1193, 0.2950]])
AFTER   mix.shape     : torch.Size([2, 441000])
tensor([[0.0486, 0.2010, 0.7882,  ..., 0.5912, 0.1193, 0.2950],
        [0.0486, 0.2010, 0.7882,  ..., 0.5912, 0.1193, 0.2950]])
```

...gives a stereo track (I think), but there is still a `ValueError` raised on the unpacking.

The `batch, channels, length` are used in the subsequent call to `torch.zero()` but I can't work out which arguments are which and how to investigate further as the arguments are passed as positional arguments and I'm unsure how `torch.zero()` unpacks them with `(*size,*)`.

So two things...

1) Is the suggested expansion of mono correct?

```python
        if mix.shape[0] == 1:
            mix = mix.repeat(2, 1)
```

2) What are the expected values of `batch`, `channels` and `length`? I think as its forcing to stereo `channels` should probably be `2` (which equates to `mix.shape[0]` and `length` is likely the length of the signal (which equates to `mix.shape[1]`) but I don't understand what `batch` should be. :thinking:

groadabike:
I am going to close this PR as this code was moved to the recipe for the challenge.

Also, I simplify the logic to work only for this challenge

PR #111, opened 2022-11-15
110 | Updating repo for flake8 in pre-commit; pre-commit autoupdate

ns-rse:
Closes #110 

Updating the `repo:` target for `flake8` in the `.pre-commit-config.yaml`. At the same time I've updated all other `rev:` of the enabled repositories/hooks.

PR #118, opened 2022-11-20
117 pre commit configyaml has outdated mypy

jonbarker68:
Updating from mypy 0.982 to 0.991 and fixing a few type hints in the code so that it continues to pass.


ns-rse (approved):
Looks good, thanks @jonbarker68 

I had tried out rev v0.991 in #111 but it broke things and was intending on updating this week as it was a bit more involved.

PR #119, opened 2022-11-21
98 | msbg evaluator nomenclature/PEP8 compliance; docstrings for mbstoi

ns-rse:
Giving the `msbg` evaluator the PEP8 treatment for variable/function/method/class names.

Realised I'd missed variables in a docstring for `mbstoi` evaluator, unsure what descriptions should be so if anyone can point me to explanations or knows what they should be that would be great, thanks.

### DIFF
--- a/clarity/evaluator/msbg/msbg_utils.py
+++ b/clarity/evaluator/msbg/msbg_utils.py
@@ -158,7 +158,7 @@ def firwin2(
 
 def fir2(
     filter_length: int,
-    sample_frequency: np.ndarray,
+    frequencies: np.ndarray,
     filter_gains: np.ndarray,
     n_interpolate: Optional[int] = None,
 ) -> Tuple[np.ndarray, int]:
@@ -168,7 +168,7 @@ def fir2(
 
     Args:
         filter_length (int): Order
-        sample_frequency (ndarray): The frequency sampling points (0 < sample_frequency < 1) where 1 is Nyquist rate.
+        frequencies (ndarray): The frequency sampling points (0 < frequencies < 1) where 1 is Nyquist rate.
                         First and last elements must be 0 and 1 respectively.
         filter_gains (ndarray): The filter gains at the frequency sampling points.
         n_interpolate (int, optional): Number of points for freq response interpolation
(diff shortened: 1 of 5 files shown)


clarity/evaluator/msbg/msbg_utils.py
```diff
@@ -105,167 +105,185 @@ def read_gtf_file(gtf_file):
...
     for key in data:
         if isinstance(data[key], list):
             data[key] = np.array(data[key])
     return data
 
 
-def firwin2(n, f, a, window=None, antisymmetric=None):  # pylint: disable=W0613
+def firwin2(
+    n_taps: int,
+    sample_frequency: np.ndarray,
```
jonbarker68, line 118: I think here, "f" needs to be "frequencies" - it defining a set of frequency positions at which the gains will be specified.  "sampling_frequency" would more likely be understood as the specific frequency at which the signal has been sampled.

clarity/evaluator/msbg/msbg_utils.py
```diff
@@ -105,167 +105,185 @@ def read_gtf_file(gtf_file):
     # Fix filename if necessary
     dirname = os.path.dirname(os.path.abspath(__file__))
     gtf_file = os.path.join(dirname, gtf_file)
-    with open(gtf_file, "r", encoding="utf-8") as fp:
-        data = json.load(fp)
```
jonbarker68, line 109: I'm a bit conflicted about this one: I've seen 'fp' used for "file pointer" in these short lived contexts in lots of pro code - I feel it is more conventional than "file_pointer" which seems a bit cumbersome.  If we have multiple files open we'd want to name the pointers e.g. "input_fp" and "output_fp" ... this would be more readable to me than "input_file_pointer" and "output_file_pointer".   

Unless you feel strongly otherwise, I suggest reverting to 'fp' but adding  "fp" to the "good-names' list in .pylintrc -- we are at least accepting it actively and documenting it as a list of short names we allow..

jonbarker68:
Thanks for this! The first point in my review is important; the 2nd more a matter of opinion... I kind of feel that "fp" is more conventional, regardless of PEP8

clarity/evaluator/msbg/msbg_utils.py
ns-rse: Done and thanks for the pointer to `good-names` in `.pylintrc`, not something I was aware of but will certainly be leveraging from now on.

PR #128, opened 2022-11-23
Spacing between logos and coloured Cadenza to show in dark mode

jonbarker68:
Fixed spacing between the logos.

The Cadenza logo looked terrible in dark mode - black on black. I've coloured it now so it shows against black or white. But the colour is not good :(   We need a graphic designer to improve this :-)


ns-rse (approved):
Thanks for correcting this @jonbarker68 I know next to nothing about graphic design I'm afraid.

groadabike:
How the logo looks if we add a white background instead of a transparent PNG?

groadabike:
My expert says we could use a 30% or 40% grey as is a colour that doesn't vibrate with black or white.
Do we have a template for the logo? I could give that a try.

But, the logo seems to be too complex, so, they may need a background so it always looks the same.
For example, the coat of arms in the Uni logo. It always has a colour (white over purple now) or a frame (in that black and white version)
https://www.sheffield.ac.uk/brand-toolkit/logo

PR #143, opened 2022-12-02
Ns rse/141 missing json

### LINKED ISSUE #141: The pip install method does not install msbg_hparams into the msbg folder
As described in the title, when installing pyclarity using pip, the installed msbg folder does not include msbg_hparams that include important json files for the cochlea simulation, and therefore prevents to use any auditory loss simulation since this folder is hard-coded in cochlea.py (line 244). 

I tested with both 0.1.0 and 0.2.0 versions and reproduced the bug each time.

When installing with the git method, the msbg_hparams is included as expected.

Could you fix this in the pipy installer ? 

Best regards,

ns-rse:
Closes #141 

Updates `setup.cfg` to include `*.json` under `clarity/` package directory and at the same time includes `recipes/**/*.json` and `recipes/**/*.csv` files too.

Also took the opportunity to `.gitignore` all `*.wav` files that I inadvertently added.

Once this is merged the `main` branch can be tagged with `v0.2.1` and a release to PyPI will be triggered.

Fixing `v0.1.0` wil require making a specific branch from the commit tagged with `v0.1.0` and `git cherry-pick` this commit over to include the changes. Once incorporated the commit on that branch and a release can be made to tag it should trigger a release of `v0.1.1`.


recipes/cec2/baseline/exp/exp/si.csv
```diff
@@ -0,0 +1,151 @@
+scene,listener,haspi
```
jonbarker68, line 1: This file exp/exp/si.csv shouldn't be included, there is already an exp/si.csv (worrying though because the numbers are different)

recipes/cec2/baseline/exp/exp/si_unproc.csv
```diff
@@ -0,0 +1,151 @@
+scene,listener,haspi
```
jonbarker68, line 1: This file exp/exp/si_unproc.csv shouldn't be included, there is already a exp/si_unproc.csv (worrying though because the numbers are different)

jonbarker68 (changes requested):
This all looks good but please remove the exp/exp/si.csv and exp/exp/unproc_si.csv. These files are already in exp/. I think there is actually a config error in the recipe that ends up putting results in exp/exp rather than exp/. but that's a separate issue and not a big problem)

As an aside: I note that the results in exp/exp/ don't match the existing one's exp/ . This worried me so I have rerun to check and the ones in exp/ appear correct.  So, yes, please delete the CEC2/baseline/exp/exp files.

jonbarker68 (approved):
Thanks. This all looks good.

ns-rse:
Thanks for the sanity check @jonbarker68 I did wonder why there were two similar files but they're only there on my local branch which is mildly confusing as I branched anew from `main`.

Anyway, I'll merge this release `v0.2.1` and then sort out a `v0.1.1` release too.

PR #161, opened 2023-01-09
Added a-to-b conversion in HOA_Tools_cec2.py

jwillbailey:
Adds A-to-B conversion so that ICASSP recordings can be processed using clarity tools pipeline

### DIFF
--- a/clarity/data/HOA_tools_cec2.py
+++ b/clarity/data/HOA_tools_cec2.py
@@ -11,10 +11,35 @@ from scipy.special import comb
 
 logger = logging.getLogger(__name__)
 
+def convert_a_to_b_format(flu: np.ndarray, frd: np.ndarray, bld: np.ndarray, bru: np.ndarray):
+    """Converts 1st order A format audio into 1st order B format
+
+    Args:
+        flu (np.ndarray): Front-left-up audio
+        frd (np.ndarray): Front-left-up audio
+        bld (np.ndarray): Front-left-up audio
+        bru (np.ndarray): Front-left-up audio
+
+    Raises:
+        TypeError: input must be numpy array
+        ValueError: all inputs must have same dimensions
+
+    Returns:
+        nd.array: 4xN array containing B-format audio. indexed w,x,y,z 
+    """
+    
+    shapes = [flu.shape, frd.shape, bld.shape, bru.shape]
+    if not all(shape == shapes[0] for shape in shapes):
+        raise ValueError("All inputs need to have same dimensions")
+        
+    w = flu+frd+bld+bru
+    x = (flu-bld)+(frd-bru)
+    y = (flu-bru)-(frd-bld)
+    z = (flu-bld)+(bru-frd)
+
+    return np.stack([w,x,y,z])
 
 # Code for generation ambisonic rotation matrices
-
-
 @njit
 def compute_rotation_matrix(n: int, foa_rotmat: np.ndarray) -> np.ndarray:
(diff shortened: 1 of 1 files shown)


clarity/data/HOA_tools_cec2.py
```diff
@@ -12,9 +12,38 @@
...
 
-# Code for generation ambisonic rotation matrices
+def convert_a_to_b_format(
+    flu: np.ndarray, frd: np.ndarray, bld: np.ndarray, bru: np.ndarray
+):
+    """Converts 1st order A format audio into 1st order B format
+
+    Args:
+        flu (np.ndarray): Front-left-up audio
+        frd (np.ndarray): Front-left-up audio
```
jonbarker68, line 22: front-right-down

clarity/data/HOA_tools_cec2.py
```diff
@@ -12,9 +12,38 @@
...
+        bld (np.ndarray): Front-left-up audio
```
jonbarker68, line 23: back-left-down

clarity/data/HOA_tools_cec2.py
```diff
@@ -12,9 +12,38 @@
...
+        bru (np.ndarray): Front-left-up audio
```
jonbarker68, line 24: back-right-up

jonbarker68 (changes requested):
Thanks.  Just needs the Args annotations fixing

jonbarker68 (approved):
Looks good. Thanks

clarity/data/HOA_tools_cec2.py
```diff
@@ -12,9 +12,38 @@
...
+
+    Returns:
+        nd.array: 4xN array containing B-format audio. indexed w,x,y,z
+    """
 
+    shapes = [flu.shape, frd.shape, bld.shape, bru.shape]
+    if not all(shape == shapes[0] for shape in shapes):
+        raise ValueError("All inputs need to have same dimensions")
+
+    w = flu + frd + bld + bru
```
ns-rse, line 38: Does the different ways in which the arrays are added have any meaningful names that could be used?

`w` doesn't say anything about how they have been arranged for this instance it could be `sum_all` as they are all added. Not sure if there is anything more meaningful for this or the others but would be easier than refactoring at a subsequent time.

ns-rse (commented):
The actions are really simple but it would be useful to help improve/maintain coverage if a unit-test were written for this new function.

Let me know if unsure about unit tests, happy to help.

clarity/data/HOA_tools_cec2.py
```diff
@@ -12,9 +12,38 @@
 logger = logging.getLogger(__name__)
 
 
-# Code for generation ambisonic rotation matrices
+def convert_a_to_b_format(
+    flu: np.ndarray, frd: np.ndarray, bld: np.ndarray, bru: np.ndarray
+):
+    """Converts 1st order A format audio into 1st order B format
```
ns-rse, line 18: In a similar vein does "1st order A format" and "1st order B format" have any more descriptive meaning that someone (e.g. me!) unfamiliar with what is actually being done might get a little more information from?

jwillbailey:
Hi Niel

Sorry, these are technical terms which refer to the spatial encoding of the soundfield that is recorded by the 3D microphone.

A-format is the signals produced by a tetrahedral array of cardioid (directional) sensors and B-Format is a spatial decomposition into the spherical Fourier domain which allows for various transformations.

I can add a URL to some literature which explains and defines these terms?

Will

PR #162, opened 2023-01-09
Haaqi PEP8 nomenclature

ns-rse:
Another step along the renaming path, this time for the `clarity.evaluator.haaqu.haaqi` sub-module.

A lot smaller this one!

### DIFF
--- a/clarity/evaluator/haaqi/haaqi.py
+++ b/clarity/evaluator/haaqi/haaqi.py
@@ -31,12 +31,12 @@ def haaqi_v1(
     Arguments:
     reference (ndarray):  Clear input reference speech signal with no noise or distortion.
             If a hearing loss is specified, NAL-R equalization is optional
-    reference_freq (int): Sampling rate in Hz for signal x
+    reference_freq (int): Sampling rate in Hz for reference signal.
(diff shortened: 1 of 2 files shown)


clarity/evaluator/haaqi/haaqi.py
```diff
@@ -50,73 +60,90 @@ def haaqi_v1(
 
     # Auditory model for quality
     # Reference is no processing or NAL-R, impaired hearing
-    xenv, xbm, yenv, ybm, xsl, ysl, fsamp = eb.ear_model(x, fx, y, fy, hl, eq, level1)
+    (
+        reference_db,
+        xbmreference_basilar_membrane,
```
jonbarker68, line 65: I think xbmreference_basilar_membrane was meant to be just 'reference_basilar_membrane'

jonbarker68 (changes requested):
Looks good. Just one small typo with a variable name. Thanks.

clarity/evaluator/haaqi/haaqi.py
ns-rse: Thanks for catching that @jonbarker68, corrected and I also noticed I'd not correctly updated the docstrings to use the new variable names.

Issue #164, opened 2023-01-12
NAL-R default parameters


groadabike:
The default value for the CFS parameters is 

https://github.com/claritychallenge/clarity/blob/14e21eff76a1350cabaa19ad26f2556892d51761/clarity/enhancer/nalr.py#L156-L158

and it should be

``` 
if cfs is None: 
     cfs = np.array([250, 500, 1000, 2000, 4000, 6000]) 

```

PR #165, opened 2023-01-12
Fix default center frequencies

groadabike:
Corrects #164 

Signed-off-by: Gerardo Roa Dabike <gerardo.roa@gmail.com>


clarity/enhancer/nalr.py
```diff
@@ -154,7 +154,7 @@ def build(
             delay
         """
         if cfs is None:
-            cfs = np.array([250, 500, 1000, 2000, 3000, 6000])
+            cfs = np.array([250, 500, 1000, 2000, 4000, 6000])
```
jonbarker68, line 157: These default cfs are the audiometric cfs stored in the constructor, i.e. there is no need to interpolate the hl value, so the correct implementation would be

```python
if cfs is not None:
    hl =  self.hl_interp(np.array(hl), np.array(cfs))
```

jonbarker68 (changes requested):
Hi, thnaks but this is not really the right fix.

If the user is not providing new cfs then the code should just be skipping the interpolation stage and using the HL values directly. Fixing it to define cfs that match the internal ones and then interpolating will just perform a redundant interpolation. 

See preferred solution below.

jonbarker68 (approved):
This looks good now. Thanks!

PR #168, opened 2023-01-19
Switching to pre-commit.ci

### LINKED ISSUE #166: Switch to pre-commit.ci
Currently there is a `pre-commit` GitHub Action in place which runs `pre-commit` against Pull Requests. However this sometimes fails if code hasn't been linted correctly locally first (see [example](https://github.com/claritychallenge/clarity/actions/runs/3911430678/jobs/6684813513)).

One way of getting around this is to use [pre-commit.ci](https://pre-commit.ci) which can be configured to commit and make the changes to the Pull Request directly (e.g. if `black` is applied to code).

This shortens the feedback loop and lessens the burden on contributors.

ns-rse:
Closes #166

Removing `.github/workflows/lint.yaml` and have configured [pre-commit.ci](https://pre-commit.ci) to run `pre-commit` against PRs instead.

This has a number of advantages...

* It is faster as it uses cached environments for the hooks that are to be run rather than installing the dependencies and downloading each hook each time `.github/workflows/lint.yaml` is run.
* Changes made to files by tools such as `black` and some of the `pre-commit` hooks such as `trailing-whitespace` are commited and pushed back to the PR without the CI simply failing as is currently the case. This can speed up the development process.
* `pre-commit.ci` will also periodically (currently `weekly`) update the `rev` version of the configured hooks.


ns-rse:
I think GitHub Actions may be confused by the removal of `.github/workflows/lint.yaml`.

If happy with this switch we can over-ride the requirement those tests to pass.

ns-rse:
For reference an example of where this would be beneficial because `pre-commit` failed and required minor changes is [this job](https://github.com/claritychallenge/clarity/actions/runs/3948963341/jobs/6759596990), where I [manually corrected the linting error](https://github.com/claritychallenge/clarity/pull/116/commits/96bc00bb8d921e09c6ff8a0cd7c815b002617c9f) (only to discover other issues with the [tests failing]() but that is a separate issue).

PR #172, opened 2023-01-30
PEP8 nomenclature for some HOA_tools_cec2.py; improving some tests

ns-rse:
Further work on PEP8 nomenclature, starting to address the `clarity/data/` modules.

* PEP8 nomenclature for some HOA_tools_cec2.py and improving some tests. Couldn't work out what some variables are though, happy to take feedback in this PR and update.
* Renaming a few variables and tidying up a bit in cec1.
* Docstrings in demo_data.py
* Also adding check_yaml to .pre-commit-config.yaml
* Docstrings and typing for scene_builder_cec2.py

### DIFF
--- a/clarity/data/HOA_tools_cec2.py
+++ b/clarity/data/HOA_tools_cec2.py
@@ -97,21 +97,21 @@ def compute_rotation_matrix(n: int, foa_rotmat: np.ndarray) -> np.ndarray:
 
 
 @njit
-def centred_element(r: np.ndarray, i: int, j: int):
+def centred_element(reference: np.ndarray, row: int, col: int):
     """Get value from centered element indexing.
 
     Args:
-        r (matrix): input matrix
-        i (int): row index
-        j (int): column index
+        reference (matrix): reference input matrix
+        row (int): row index
+        col (int): column index
 
     Returns:
         Any: matrix element
     """
-    offset = int((r.shape[0] - 1) / 2)
-    if i > r.shape[0] or j > r.shape[1]:
+    offset = int((reference.shape[0] - 1) / 2)
+    if row > reference.shape[0] or col > reference.shape[1]:
         raise IndexError
-    output = r[i + offset, j + offset]
+    output = reference[row + offset, col + offset]
     return output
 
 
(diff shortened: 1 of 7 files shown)


clarity/data/HOA_tools_cec2.py
```diff
@@ -145,74 +145,80 @@ def P(i, a, b, el, r):
 
 
 @njit
-def U(m, n, el, r):
+def U(degree, n, order, rotation_matrices):
     """U coefficient initialiser for rotation matrix calculation.
 
     Args:
-        m (int): degree
```
jwillbailey, line 152: the use of degree here is not as in subdivision of a circle, but the parameters of a spherical harmonic decomposition.

The variable names for the code called by the rotation matrix generation are defined here:

 'Rotation Matrices for Real Spherical Harmonics. Direct Determination
    by Recursion' Joseph Ivanic and Klaus Ruedenberg J. Phys. Chem. 1996, 100, 15,
    6342–6347

and 

'Forier Acoustics' E. G. Williams https://doi.org/10.1016/B978-0-12-753960-7.X5000-1

in this cale, m refers ro the upper parameter of spherical harmonic component Y and n the lower.

clarity/data/HOA_tools_cec2.py
```diff
@@ -145,74 +145,80 @@ def P(i, a, b, el, r):
...
 
 
 @njit
-def V(m, n, el, r):
+def V(degree, n, order, rotation_matrices):
     """V coefficient initialiser for rotation matrix calculation.
 
     Args:
-        m (int): degree
+        degree (int): degree valid values are 0, 1 and -1.
```
jwillbailey, line 168: the valid values for this would only be -1, 0, 1 for the case of the first order rotation submatrix.
for higher orders, which this code computes, valid inputs are int(|m|)<=el

clarity/data/HOA_tools_cec2.py
```diff
@@ -384,31 +391,38 @@ def binaural_mixdown(ambisonic_signals, hrir, hrir_metadata):
...
-    """Convolve HOAIRs with signals.
+def ambisonic_convolve(
+    signal: np.ndarray, hoa_impulse_responses: np.ndarray, order: int
+) -> np.ndarray:
+    """Convolve HOA Impulse Responses with signals.
 
     Args:
         signal (ndarray[samples]): the signal to convole
-        ir (ndarray[samples, channels]): the HOA impulse responses
+        hoa_impules_response (ndarray[samples, channels]): the HOA impulse responses
```
jwillbailey, line 410: typo here. should be impulse

clarity/data/HOA_tools_cec2.py
ns-rse: Thanks @jwillbailey that maths there is a bit beyond me, I've changed the text to reflect what you've written and modified docstrings to correctly cite the articles in Restructured Text. In doing so I looked up the DOI for the papers and found there was a [correction to the Ivanic and Klaus paper](https://pubs.acs.org/doi/10.1021/jp9833350), not sure if it would have any bearing on what has been implemented here. :shrug:

jwillbailey:
Yes, this python code is actually a translation of part of the google resonance code<https://github.com/resonance-audio/resonance-audio/blob/master/resonance_audio/ambisonics/hoa_rotator.cc>. The correction for the original paper was taken into account.

I agree this stuff is pretty dense and not very readable. All worked out by much cleverer heads than mine.

PR #183, opened 2023-02-22
Adding local hook .pre-commit-config.yaml to run pylint

### LINKED ISSUE #182: Enable pylint in pre-commit checks
We have linted the code base using [pylint](https://pylint.pycqa.org/en/latest/) (see issue #42 and PR #43) however we do not enforce `pylint` checks against new commits.

We can do this by incorporating a `pylint` hook into [`pre-commit`](https://pre-commit.com) which will carry through to the [`pre-commit.ci`](https://pre-commit.ci) GitHub Action.

In instances like this we should use a `local` hook because using the one provisioned by the [PyCQA](https://github.com/PyCQA/pylint) will not include the necessary packages required to perform the checks. This can be achieved using the following snippet which will use the existing `.pylintrc` configuration.

```yaml
  - repo: local
    hooks:
      - id: pylint
        args: ["--rcfile=.pylintrc"]
        name: Pylint
        entry: python -m pylint
        language: system
        files: \.py$
```

ns-rse:
Closes #182

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -44,3 +44,6 @@ jobs:
           coverage xml
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v3
+      - name: pylint
+        run: |
+          pylint --rcfile=.pylintrc
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -40,7 +40,7 @@ repos:
         args: ["--profile", "black"]
 
   - repo: https://github.com/nbQA-dev/nbQA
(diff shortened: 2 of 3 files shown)


ns-rse:
Wasn't quite as simple as I'd hoped, have to disable the `local` hook for `pylint` under `pre-commit.ci` since it uses cached environments (i.e. the hook provided by PyCQA for `pylint`) which will _always_ fail since it doesn't include all the packages Clarity depends on.

I've therefore moved running `pylint` to the GitHub Action `.github/workflows/run_test.yaml` where it will run (after we've fixed the broken tests!).

PR #193, opened 2023-03-02
Demo data

groadabike:
Add demo data so participants can start playing with recipe before downloading all the data

### DIFF
--- a/recipes/cad1/task1/baseline/README.md
+++ b/recipes/cad1/task1/baseline/README.md
@@ -23,6 +23,7 @@ and `cadenza_cad1_task1_augmentation_fma_small.tar.gz`.
 
 Unpack packages under the same root directory using
 
+
 ```bash
 tar -xvzf <PACKAGE_NAME>
 ```
@@ -99,6 +100,27 @@ cadenza_data
             ├───001
             ├───...
 ```
+
+### 1.3 Demo data
+
+To help you to start with the challenge, we provide a small subset of the data.
+the `demo_data` folder contains a single song and two listeners from the validation set.
+
+To use the demo data, simply download the package `cadenza_data_demo.tar.xz`
+from [here](https://drive.google.com/drive/folders/1Yxo_R-yPByEUvX5O5lhsHk3tW1ek5qKW?usp=share_link)
+and unpack it under `recipes/cad1/task1/`, i.e., one level above the baseline directory.
+Note that the `root.path` variable in `config.yaml` is already set to the demo data by default.
+
+To unpacked the demo data, run:
+
+```bash
+tar -xvf cadenza_data_demo.tar.xz
+```
+
+
+```text
+
+
 
 ## 2. Baseline
 
--- a/recipes/cad1/task1/baseline/enhance.py
+++ b/recipes/cad1/task1/baseline/enhance.py
@@ -138,6 +138,7 @@ def map_to_dict(sources: np.ndarray, sources_list: List[str]) -> Dict:
     return signal_stems
 
 
+# pylint: disable=unused-argument
 def decompose_signal(
     model: torch.nn.Module,
     signal: np.ndarray,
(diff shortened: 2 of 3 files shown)


recipes/cad1/task1/baseline/README.md
```diff
@@ -100,6 +101,27 @@ cadenza_data
             ├───...
 ```
 
+### 1.3 Demo data
+
+To help you to start with the challenge, we provide a small subset of the data.
+the `demo_data` folder contains a single song and two listeners from the validation set.
```
ns-rse, line 107:
```suggestion
The `demo_data` folder contains a single song and two listeners from the validation set.
```

recipes/cad1/task1/baseline/README.md
```diff
@@ -100,6 +101,27 @@ cadenza_data
...
+
+To use the demo data, simply download the package `cadenza_data_demo.tar.xz`
+from [here](https://drive.google.com/drive/folders/1Yxo_R-yPByEUvX5O5lhsHk3tW1ek5qKW?usp=share_link)
+and unpack it under `recipes/cad1/task1/`, i.e., one level above the baseline directory.
+Note that the `root.path` variable in `config.yaml` is already set to the demo data by default.
+
+To unpacked the demo data, run:
```
ns-rse, line 114:
```suggestion
To unpack the demo data, run:
```

recipes/cad1/task1/baseline/enhance.py
```diff
@@ -138,6 +138,7 @@ def map_to_dict(sources: np.ndarray, sources_list: List[str]) -> Dict:
     return signal_stems
 
 
+# pylint: disable=unused-argument
```
ns-rse, line 141: The `left_audiogram` and `right_audiogram` are the cause of this linting error. Since they don't appear to be used in this function hey can be removed from the `def()` (and docstring).

ns-rse (changes requested):
All coming together nicely @groadabike :+1: 

Some minor typos, but also the linting error can be remedied rather than disabled.

recipes/cad1/task1/baseline/enhance.py
groadabike: This was intentionally left in this way. The idea is to show participants that even thou the baseline is not using the audiograms, they can modify the function and take use of them.

groadabike (commented):
The idea of the baseline is to offer a system where participants can change just one part and still have a full system running. The left and right audiograms are left in that function so they can use them when changing that function without the need of changing the baseline code in many places.

I added the pylint ignore because I didn't find a more elegant way to keep those parameters in the function.

jonbarker68 (approved):
All looks good.

jonbarker68 (approved):
Looks good. 👍

ns-rse (approved):
Sorry I was teaching this morning and unable to review until now.

Approving but I think the `decompose_signal()` options could be modified to be a little more helpful to participants.

groadabike:
Thank you @ns-rse 
Can you please send me later the modification we should do?

ns-rse:
Sure I left them in the [comment](https://github.com/claritychallenge/clarity/pull/193#discussion_r1124541780).

Copy and pasting below....

Fair enough, presumably anyone working with the code at this level would know how to modify the function?

I think it might be possible to make life a little easier for users though and have two possible suggestions...

## Allow options and optionally replace `signal`

If a user want to process left or right audograms you could...

```python
    if left_audiogram is not None:
        signal = left_audiogram
    elif right_audiogram is not None:
        signal = right_audiogram
    elif left_audiogram is not None and right_audiogram is not None:
        Exception("Please specify on one of left_audiogram or right_audiogram")
```

...and note in the docstring that if either `left_audiogram` or `right_audiogram` is specified it will replace `signal`.

## Remove the options and improve the docstring

Alternatively just remove the  `left_audiogram` and `right_audiogram` options and improve the docstring to reflect the fact that `signal` can be a combined channel or a single decomposed channel of either left or right audio.


Either of these would I think make it easier for participants to use the function without having to modify it.

PR #200, opened 2023-03-06
Fixing type error for haspi

jonbarker68:


clarity/evaluator/haspi/ip.py
```diff
@@ -210,7 +210,7 @@ def get_neural_net():
 
 def nn_feed_forward_ensemble(
     data: np.ndarray, neural_net_params: Dict, weights_hidden, weights_out
-) -> float:
+) -> np.ndarray:
```
jonbarker68, line 213: This is an error. The nn returns an nd.array containing the activations of the output layer

jonbarker68:
This is only failing the same test that always fails.  I'm going to merge this without review because I need the fix to make progress with the CPC2 recipes that are already overdue :-(

PR #208, opened 2023-03-11
Jpb/204 de duplicate files

jonbarker68:
These have now been dealt with by variously,
I) deleting copies of files and making code edits so the remaining copy of the file is referenced
ii) editing a copy of the duplicated file to introduce an explicit difference (e.g. in a header comment) - this is useful where there is good reason to have a duplicate copy, e.g. in recipes that are best maintained independently of each other.


ns-rse (approved):
Useful de-cluttering :+1:

PR #212, opened 2023-03-12
Jpb/203 speed up test suite by reducing length of samples

jonbarker68:
Work on the test-suite to speed up all the slowest tests


jonbarker68:
I tried to speed up the GitHub action by using caching, as explained [here](https://github.com/actions/setup-python#caching-packages-dependencies) but it only slowed things down (i.e. took longer to restore from cache than download things from PyPI, and only seems to cache the download not the full setup). But I've left the commented out lines in the workflows/run_tests.yml file for reference.

ns-rse (approved):
Great work cutting down the length of sampels, thanks @jonbarker68

PR #214, opened 2023-03-12
213 cadenza add batches evaluation in cad1 task1

groadabike:
Add evaluation in batches for Task1 to match behaviour implemented in task2

### DIFF
--- a/recipes/cad1/README.md
+++ b/recipes/cad1/README.md
@@ -10,10 +10,10 @@ and the instruction on how to obtain the data.
 
 [Learning Resources](http://cadenzachallenge.org/docs/learning_resources/learning_intro) are available on the challenge homepage. Specific topics include...
 
-* [Installation](https://colab.research.google.com/drive/1m2EvNyBmaZT5njbWkHOUAIvPKyxR8iRc?usp=sharing)
-* [Interacting with Metadata](https://colab.research.google.com/drive/1XtoMp8Vv_6ZtU4AE9OmQj7uRbsCzw_ZY?usp=sharing)
-* [Baseline system CAD1 - Task1](https://colab.research.google.com/drive/1j8L2dBo8rTQsbKzPTxAymVvDMgHtgtnK?usp=sharing)
-* [Baseline system CAD1 - Task2](https://colab.research.google.com/drive/1fgvnrgCUiylUHL26ElShp6egIZiU_vNz?usp=sharing)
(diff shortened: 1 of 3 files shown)


recipes/cad1/task1/baseline/merche_batches_results.py
```diff
@@ -0,0 +1,25 @@
...
+import pandas as pd
+from omegaconf import DictConfig
+
+
+@hydra.main(config_path="", config_name="config")
+def join_batches(config: DictConfig) -> None:
+    """
+    Join batches scores into a single file.
+
+    """
```
ns-rse, line 13:
```suggestion
    Arguments:
        config (dict): Dictionary of configuration options. The '.evaluate.batch_size' is extracted to determine how many batches there are to combine.
    """
```

recipes/cad1/task1/baseline/merche_batches_results.py
```diff
@@ -0,0 +1,25 @@
...
+    batches_results = []
+    for batch in range(config.evaluate.batch_size):
+        batches_results.append(
+            pd.read_csv(f"scores_{batch + 1}-{config.evaluate.batch_size}.csv")
+        )
+    df_res = pd.concat(batches_results, ignore_index=True)
+    df_res.to_csv("scores.csv", index=False)
```
ns-rse, line 20: Would it perhaps be useful to have an additional argument `outpath: Union[str, Path] = "./"` and make the output location configurable? I'm not sure how/where people use this but making such things flexible helps users.

Would also require...

```python
    outpath = Path(outpath)
```

...to ensure the object is of type `Path()` and could then...

```suggestion
    df_res.to_csv(outpath / "scores.csv", index=False)
```

ns-rse (changes requested):
Perhaps a typo in the filename `merche_batches_results.py` should it perhaps be `merge_batch_results.py`?

Made a suggestion to make the output perhaps more flexible, but am unsure if it would fit with workflow. I don't think including it would cause any problems though.

recipes/cad1/task1/baseline/merche_batches_results.py
groadabike: The output directory is already defined in `path.exp_folder` in the config.
So, the combined file will be stored next to the batches results in the experiment directory.
That is one of the beauties of Hydra, you can define the experiment path and that path is then implicit throughout the whole experiment.

ns-rse (approved):
Thanks @groadabike looks ready to merge to me.

PR #221, opened 2023-03-14
jpb/220-enable-some-pylint-error-checking-in-ci

jonbarker68:
Adding pylint error checking into the ci part of the precommits. The .pre-commit-config.yaml now has two entries for pylint: one which runs errors only which is compatible with ci, and one that runs as 'repo: local' and uses the .pylintrc which will pick on all the usual style warnings etc, but which is skipped over by the ci.

I have also run pylint with the errors-only flag over the full repo and fixed all issues that were picked up - or added pylint ignore against the ones that appeared to be false-positives.

[Ruff](https://github.com/charliermarsh/ruff): I have also added 'ruff' to the precommit hooks. 'ruff' is a rust-based linter that is super fast (but doesn't yet have quite as many rules as pylint). I fixed a number of issues that the ruff checked picked up on. Ruff might eventually be able to replace pylint but for now we can just keep it in as an extra linter because it takes very little time to run.


jonbarker68:
@groadabike  - there is nothing significant in this PR - mostly just fixing line-too-long type errors and issues that pylint considers 'errors' i.e Exxx codes. Could you please give it a quick review (I don't want to get in the habit of overriding the need for reviews :-).  Once this is integrated pylint will pass with the '-E'  i.e. errors only flag which can then be included safely in the precommit ci path.

jonbarker68:
Going to merge this PR if

groadabike (approved):
Looks good to me.
Great job dealing with these pylint errors

PR #224, opened 2023-03-15
222 fix issues raised by pylint

jonbarker68:


jonbarker68:
Lots of files have been touched but most changes are very trivial. The commit has add a .pylintrc.precommit.ini that is used as a pylint configuration for pre-commit. This has a relatively small list of checks disabled - mostly of the 'Refactor' type. The existing code base should now pass with this configuration.  The previous .pylintrc is still present and developers should be encouraged to use it as the 'R' warnings are still useful.

The codecov/patch is failing simply because many of the minor edits are in code that does not have existing test coverage. We have a separate set of issues raised for dealing with this.

ns-rse:
Sorry for the lack of activity from me recently, I've been on strike.

Great work going through all these @jonbarker68 

I've taken the liberty of adding the commits to `.git-blame-ignore-revs` under an appropriate heading along with the hashes because these are minor changes. This means the `git blame` won't be attributed to you (conditional on the config, see below). I've _included_ one (3a9a31) which introduced pytorch<2.0 to the dependencies. This has been committed and pushed.

In order to get the list of hashes I used the following `git log` to restrict to branch/date and just print the hashes (although I checked first the commit messages).

```
git log  main..222-fix-issues-raised-by-pylint --after="2023-03-14" --pretty=format:"%H # %s"
```

GitHub recognises these `.git-blame-ignore-revs` file automatically so they won't be displayed here on GitHub, but locally you may want to run the following so that they are also ignored...

```
git config blame.ignoreRevsFile .git-blame-ignore-revs
```

...or add the following to `.git/config` in the repository...

```
[blame]
    ignoreRevsFile = .git-blame-ignore-revs
```

I wrote a blog post on this at [Who's to blame?](https://rse.shef.ac.uk/blog/git-blame/).

ns-rse (approved):
`pylint` is happy and tests pass so this is good to go.

jonbarker68:
Thanks for checking this and explaining the git-blame process.

PR #227, opened 2023-03-16
Restricting Pytorch to version < 2.0

groadabike:
Pytorch 2.0 breaks some tests.
Restrict the version until we are sure what are the main differences between Pytorch 1 and 2 that affect the current tests.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -48,7 +48,7 @@ install_requires =
   scikit-learn>=1.0.2
   scipy>=1.7.3
   SoundFile>=0.10.3.post1
-  torch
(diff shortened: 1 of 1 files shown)


groadabike:
@jonbarker68 @ns-rse 
Changing the config to use `torch<2.0` did the trick.
The tests are not failing now.

We may need to dig a little bit to understand where the differences originate to make our recipes reproducible.
As the tests showed, the results are not the same when seeding and using Pytorch 1 vs Pytorch 2

jonbarker68 (approved):

PR #237, opened 2023-03-18
85-unit-tests-for-evaluator-module

jonbarker68:


.pre-commit-config.yaml
```diff
@@ -82,11 +82,12 @@ repos:
   - repo: local
     hooks:
       - id: pylint_warnings
-        args: ["--rcfile=.pylintrc"]
+        args: ["--rcfile=.pylintrc.precommit.ini"]
         name: pylint-warnings
         entry: python -m pylint
         language: system
         files: \.py$
+        exclude: clarity/_version.py # 3rd party code that doesn't conform
```
ns-rse, line 90: I'd hoped that #186 would help here but `setuptools_scm` doesn't (yet^1) produce a Black compliant `_version.py` as it uses `'` rather than `"` see [setuptools_scm 482](https://github.com/pypa/setuptools_scm/issues/482))

^1 I had a brief attempt at addressing that but its more involved than I thought, will get back to it somewhen.

clarity/evaluator/haspi/ip.py
```diff
@@ -4,7 +4,7 @@
 import numpy as np
 
 
-def get_neural_net() -> Tuple[Dict, List[np.ndarray], List[np.ndarray], float]:
+def get_neural_net() -> tuple[dict, list[np.ndarray], list[np.ndarray], float]:
```
groadabike, line 7: why change from `Tuple` to `tuple`, `Dict` to `dict` and so on?
Is it related to `pyupgrade`?

We may need to do this change throughout the whole repository. Including recipes.
Do we need to remove the importing of `typing` and add the `from __future__ import annotations`?

jonbarker68: Tuple, Dict etc were a hack for early versions of Python when type hints were first being  introduced. 

In modern python versions types are now built into the language as tuple, dict etc. So you don't need to import from Typing anymore.  And you can still use this in older versions of python as long as you include "from __future__ import annotations" at the start of the code.

Note, the modern type hints also allow things line "float | np.ndarray" rather than the clumpy "Union[float, np.ndarray]" and for "Optional" you can use e.g. "int | None" -- it's all much cleaner.

It's related to pyupgrade in the sense that if you do write hints in the obsolete style then pyupgrade will automatically fix them for you.

Yes, it has been changed throughout the repository and recipes.  pyupgrade has done that for us. :-)

(Note, pyupgrade will make it easy for us to keep the repo up to date. We just need to bump up the version number on its command line as we drop support for older (unsafe versions of Python) -- e.g. security fixes are no longer being made for 3.6 and will be dropped for 3.7 in a couple of months. i.e. we will shortly remove 3.7 support and hence be able to use features from 3.8 onwards)

groadabike: Thank you so much for the comprehensive message. Very clear.

It is great to know that  `pyupgrade` did all this tedious work for us.

jonbarker68:
@ns-rse, @groadabike  - this PR is for issue #237 that's adding tests for all the evaluator/ modules. I have handled all the HASPI/HAAQI/HASQI and MBSTOI  but not yet the MSBG. But I thought it best to have this reviewed and merged first so that it can be used... and then merge again when MSBG is complete. (MSBG is a lot of code an not a priority given that we're not using it in current challenges).

Regarding the tests: they are mostly basically function-level regression tests using random input.  Although input signals are random, I've tried to tune them such that the functions are operating correctly. This is important because in many tests, if the input is wrong 0's will appear in the functions and although the coverage might look good, the output won't be sensitive to changes in the code, eg.  f1() + 0 * f2() will produce the same output regardless of the content of f2(). I've also tried to use the smallest meaningful signals to keep things fast. All 60 tests run in about 5 seconds.

Note also that the PR includes some changes to the code being tested. This is mostly just typehints being introduced and a few trivial things (e.g. ugly formatting caused by line-end comments) that were worth fixing on the spot. The only significant change is in  eb.py where I have taken advantage of the tests to vectorise code in ave_covary2.  It should run much faster ... but doubt it was the bottleneck.

clarity/evaluator/haspi/eb.py
```diff
@@ -2034,48 +2028,35 @@ def ave_covary2(
...
+    #    for n in range(nseg):
+    #        # Thresh in dB SL for including time-freq tile
+    #        if signal_rms[k, n] > threshold_db:
+    #            wsync[:, k, n] = fsync[:, k]
+    #
+    # This can be written is an efficient vectorsized form as follows:
+    wsync = np.zeros((6, n_channels, nseg))
+    mask = signal_rms > threshold_db
+    fsync3d = np.repeat(fsync[..., None], nseg, axis=2)
+    wsync[:, mask] = fsync3d[:, mask]
```
groadabike, line 2046: We should try to do this in more parts of the eb.
For sure there are many places we can improve.

groadabike (approved):
It looks great. Thank you @jonbarker68 

One question.
Why in `clarity/evaluator/mbstoi/mbstoi_utils.py` are we importing `ndarray` from numpy instead of using `np.ndarray`. This is the only file where er are doing that.

clarity/evaluator/haspi/eb.py
groadabike: After this is merge, I will check if I can find ways to improve HAAQI

jonbarker68:
> One question. Why in `clarity/evaluator/mbstoi/mbstoi_utils.py` are we importing `ndarray` from numpy instead of using `np.ndarray`. This is the only file where er are doing that.

Ah, well-spotted. This is because I used MonkeyPatch to help me insert the typehints for that module and that appears to be its preferred style. We should be consistent.

Actually, ndarray is pretty poor as a type. Both too general in that it doesn't specify the dtype (i.e. is it meant to be floats or ints), and not general enough in the sense that for many of these functions a python list, or list of lists could be safely passed if it can be coerced into a numpy array -- i.e. the code would run but we'll now get an error from mypy.

numpy has its own typing module that defines types like ArrayLike that solves this problem https://numpy.org/doc/stable/reference/typing.html but a bit beyond my experience.

I noticed when writing tests though that because the original MATLAB code is not so well documented that it was very hard to work out what these numpy arrays are allowed to be, i.e. 1D, 2D, 3D ?

There's an interesting project here that attempts to solve this https://github.com/ramonhagenaars/nptyping but I think it's a bit too bleeding-edge. We just need clearer docstrings.

jonbarker68:
Pull request has been merged but branch not deleted -- tests for the MSBG code need adding to complete this issue.

Issue #247, opened 2023-03-22
Improving type hints using MonkeyType and pytest-monkeytype


jonbarker68:
It would be great to have type hints used consistently throughout the codebase, but the cost of adding them has seemed prohibitive. However, I've found this great looking tool called [MonkeyType ](https://monkeytype.readthedocs.io/en/stable) that will do runtime analysis to generate type hint suggestions automatically and can then apply them to the code. It requires running all the code in the codebase, but there is a [pytest plugin](https://github.com/mariusvniekerk/pytest-monkeytype) that will run MonkeyType during running the testsuite, so it's super easy to use. 

It can be set up to overwrite or ignore existing suggestions, so as long as it's used with a bit of care it looks safe.  

We could have a branch to run it across all the existing tests, and then just incorporate it into the workflow when working on new tests.

(There is also DropBox's [pyannotate](https://github.com/dropbox/pyannotate/commits/masterrl) but this looks unmaintained and harder to use)

ns-rse:
This looks like a great tool and well worth applying. The plugin looks invaluable and combined with the `pre-commit` hook running `mypy` would I hope make it pretty straight-forward to add.

Issue #225, opened 2023-03-15
Refractoring HAAQI


groadabike:
Running HAAQI in full-length song is very slow.
This affects both tasks  of this first round of Cadenza

groadabike:
Hi @ns-rse, @jonbarker68 

I've been trying to speed up the HAAQI code.
It is pretty slow because they have several (around 8) full autocorrelations of very long signals (about a 3-min song).
 
In the code, we are using scipy which is not supported by numba but suppose to be faster than np.correlate for long signals.
But, I did some profiling using both and couln't find a significant difference.

We can't just add `@jit(nopython=True)` on top of this function because Numba only supports the default of numpy.correlate, i.e., correlation with mode = 'valid' and not 'full'
  
```
https://numpy.org/doc/stable/reference/generated/numpy.correlate.html
https://github.com/numba/numba/blob/07f824e49e5e2f158b79e7ba787674272baf1da0/numba/np/arraymath.py#L4199
```

I tried to overwrite the numba's numpy.correlate `overload` but I wasn't able to do it successfully.
Have you ever tried something like that?
Is it worth trying to define our own `overload` to try to accelerate the code?

```
https://numba.pydata.org/numba-doc/dev/extending/overloading-guide.html?highlight=overload
```

Thank you for your help.

groadabike:
@ns-rse, @jonbarker68 

After the release of `pyclarity 0.3.0` I want to start working on improving HAAQI.
There are many changes we can do to make it more pythonic and improve its readability and performance.

For example:
1. We can cache most of the functions. This would help in all cases when the references need to be processed more than one time. And/or,
2. We can refactor the code to an abstract class. Then `haaqi`, `haspi` and `hasqi` can implement or overwrite the methods. Some functions in `haspi.eb` are common like `ear_model` that should be a method under the abstract class, and others functions are metric specific like `eb.melcor9` which is only used by `haaqi` and only need to be a method in `haaqi` class.
3. We need to add type hints.

What are your thoughts? What else would you do to improve the metrics?

ns-rse:
These all seem like eminently sensible ideas @groadabike 

Obviously vectorising loops are going to offer good speed ups and restructuring into abstract classes are really useful.

Has your profiling covered more than just numpy vs scipy for correlation?

If not it might be worth using profiling against the code base to determine which areas might benefit most from careful refactoring.

groadabike:
The profiling I am running evaluates the time of each function when running haaqi.
[...]

PR #248, opened 2023-03-23
105-inconsistent-logic-with-parameter-shift-in-evaluatorhaspiebcenterfreq

jonbarker68:
The bug in the original HASPI code was being preserved by enforcing it in the low level centre_freq function. This has been changed so that centre_freq can now work as expected and the bug has been lifted up to the top level HASPI function where it can be clearly seen.

### DIFF
--- a/clarity/evaluator/haspi/eb.py
+++ b/clarity/evaluator/haspi/eb.py
@@ -29,7 +29,7 @@ def ear_model(
     level1,
     nchan=32,
     m_delay=1,
-    shift=0.02,
+    shift=None,
 ):
     """
     Function that implements a cochlear model that includes the middle ear,
@@ -357,9 +357,6 @@ def center_frequency(
     Translated from MATLAB to Python by Zuzanna Podwinska, March 2022.
     """
 
-    # In the Matlab code, the loop below never evaluates
-    # (but the current code was trained with this bug)
-    shift = None  # This is to keep consistency with MATLAB code
     if shift is not None:
         k = 1
         A = 165.4  # pylint: disable=invalid-name
--- a/clarity/evaluator/haspi/haspi.py
+++ b/clarity/evaluator/haspi/haspi.py
@@ -68,6 +68,8 @@ def haspi_v2(  # pylint: disable=too-many-arguments too-many-locals
         hearing_loss,
         itype,
         level1,
+        # shift=0.02 # Original HASPI code had a bug such that this shift was not used
+        shift=None,
     )
 
     # Envelope modulation features


clarity/evaluator/haspi/eb.py
```diff
@@ -29,7 +29,7 @@ def ear_model(
     level1,
     nchan=32,
     m_delay=1,
-    shift=0.02,
+    shift=None,
```
ns-rse, line 32: Would it be prudent to add an explanation to the docstring about the bug/issue? This would then be part of the API documentation that is built automatically (see [here](https://claritychallenge.org/clarity/clarity.evaluator.haspi.eb.html#clarity.evaluator.haspi.eb.ear_model)).

A comment for the argument...

```
shift (float): Basal shift of the basilar membrane length. Default is 'None' to align with the original HASPI model derivation (see above note).
```

...and/or a paragraph in the longer description that precedes `Arguments`

```
**NB** - The original HASPI model derivation included a bug where the 'shift' was meant to be set to '0.02' but was never actually applied. For discussion please refer to the discussion in `Issue #105 <https://github.com/claritychallenge/clarity/issues/105>`_ for further details.
```

ns-rse (approved):
I think this is a clearer communication of why the settings are as such and it makes more sense (to me) to have the default values in `ear_model()` defined as the intended value of `None` rather than `0.02` and then resetting it.

I've suggested adding further detail to the docstring to help inform readers as this carries through to the API documentation which is built automatically.

Not a blocker to merging though if that is felt excessive.

groadabike:
It looks good to me too.
But, we need to do a lot more work in `hasqi`, `haspi`, `haaqi`, `haspi.eb` and `haspi.ebm` modules.
They don't have type hints, we need to try to vectorise all of the for loops and perhaps refractor the files. 
For example, `eb` is a dependency for the three metrics, it makes no sense to me to have it under haspi.

ns-rse:
> It looks good to me too. But, we need to do a lot more work in `hasqi`, `haspi`, `haaqi`, `haspi.eb` and `haspi.ebm` modules. They don't have type hints, we need to try to vectorise all of the for loops and perhaps refractor the files. For example, `eb` is a dependency for the three metrics, it makes no sense to me to have it under haspi.
> 
> But this is work for another PR. I will try to deal with it under the issue Speed up HAAQI #225

@jonbarker68 has created a separate issue for typehints see #247.

clarity/evaluator/haspi/eb.py
jonbarker68: Thanks. That's a good point. I'll address this before merging.

Issue #81, opened 2022-10-13
Unit tests for data module


ns-rse:
Develop unit tests for the data module (see #80 for overview).

Specific modules within `data` are...

- [ ] `data/demo_data.py`
- [ ] `data/HOA_tools_cec2.py`
- [ ] `data/scene_builder_cec1.py`
- [ ] `data/scene_render_cec1.py`
- [ ] `data/scene_render_cec2.py`
- [ ] `data/utils.py` #88

PR #254, opened 2023-03-26
81-unit-tests-for-data-module part 1

jonbarker68:
Working towards completing the unit-tests for the data module, issue #81 

Will split this over a number of PRs and will be doing a bit of test-driven refactoring along the way.

### DIFF
--- a/tests/data/test_HOA_tools_cec2.py
+++ b/tests/data/test_HOA_tools_cec2.py
@@ -4,7 +4,8 @@ import pytest
 
 # HOARotator,; ambisonic_convolve,; binaural_mixdown,;;;
 # compute_band_rotation,;; compute_rotation_matrix,;  dot,; dot,
-from clarity.data.HOA_tools_cec2 import (
+from clarity.data.HOA_tools_cec2 import (  # compute_rotation_matrix,
+    HOARotator,
     P,
     U,
     V,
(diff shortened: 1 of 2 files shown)


jonbarker68:
- Provides complete test coverage for data/HOA_tools_cec2.py 
- Fixes issues that produce numba deprecation warnings in HOA_tools_cec2
- Fixes an issue with the random_matrix fixture that is used in test_HOA_tools_cec2

tests/data/test_HOA_tools_cec2.py
```diff
@@ -1,146 +1,251 @@
...
+    r3_matrix = make_random_matrix(seed + 20)  # ditto
     assert P(
         i,
         a,
         b,
         order,
-        rotation_matrices=[random_matrix, random_matrix, random_matrix],
-    ) == pytest.approx(expected, rel=pytest.rel_tolerance, abs=pytest.abs_tolerance)
+        rotation_matrices=TypedList([r1_matrix, r2_matrix, r3_matrix]),
+    ) == pytest.approx(expected)
```
ns-rse, line 165: Is the default relative tolerance of `1e-6` ok for this test or broadly across the whole code base? 

If so we could change those defaults in `conftest.default__config()` so they are used globally and always explicitly state the values should we ever want to change them across the code base.

Issue #222, opened 2023-03-15
Fix issues raised by pylint


jonbarker68:
`pylint` is raising a lot of issues when run against the whole code base. The pre-commits force them to be dealt with when files are touched for other reasons which cause commits to become unclear. 

Propose to deal with them all in one go. The plan will be that for each one

i/ make changes to the code if there is an obvious fix (e.g. all the line length warnings)
ii/ add a disabling comment if it is an obvious false-positive 
iii/ change the .pylintrc used by recommit if it is a larger problem that we will return to (e.g. most of the refactor R warnings)

The branch will be merged when `pre-commit run --all-files` passes completely.

jonbarker68:
Issues dealt with so far:

- R1705 - unecessary elif - 9 cases
- R1732 - could use 'with' - 26 cases
- W0621 - redefine from outer scope  - 17 cases
- W1514 - Using open without explicitly specifying an encoding - 69 cases
- W0201 - Attribute defined outside init - 39 cases
- C0200 - use enumerate - 13 cases
- W0613 - unused arg - 10 cases
- W0612 - unused var - 11 cases

Issue #219, opened 2023-03-14
Support for Python 3.11


jonbarker68:
Currently appear to be two external dependencies blocking upgrade to 3.11:  pytorch and numba

- pytorch: Nightly build of pytorch currently runs on 3.11 but I think best to wait until pytorch 2.0 release due [in early March](https://pytorch.org/get-started/pytorch-2.0/#:~:text=We%20expect%20to%20ship%20the,release%20in%20early%20March%202023).
- numba: a Python 3.11 compatible version is due this month, see [discussion here](https://github.com/numba/numba/issues/8304)

Plan to test against Python 3.11 as soon as these are released

jonbarker68:
Update:
- pytorch 2.0 is now out and supports Python 3.11. But it does seem to be leading to significant changes in test results. Needs some investigation.
- numba - sounds like it will still be another week or two

jonbarker68:
Latest numba on github is supposed to support Python 3.11

```
pip install git+https://github.com/numba/numba.git
```

But it has llvmlite v0.40.0dev0 as a prerequisite 

```
pip install git+https://github.com/numba/llvmlite@v0.40.0dev0
```

On the mac this has further issues see discussion (here)[https://github.com/cvxgrp/pymde/issues/49]

```
brew install llvm@11
LLVM_CONFIG="/opt/homebrew/Cellar/llvm@11/11.1.0_4/bin/llvm-config" arch -arm64 python3 -m pip  install git+https://github.com/numba/llvmlite@v0.40.0dev0
```

But this fails with a lot of "Undefined symbols for architecture arm64"

Might have to wait a bit longer for a numba release on PyPI

PR #257, opened 2023-03-27
186 Switch to pure pyproject.toml for package configuration

### LINKED ISSUE #186: Switch to setuptools_scm and pyproject.toml
Rather than using [versioneer](https://github.com/python-versioneer/python-versioneer) to determine versions dynamically the same functionality is provided by [setuptools_scm](https://github.com/pypa/setuptools_scm/) and is cleaner/easier to install maintain as configuration is via the projects `pyproject.toml`.

At the same time switching over to [`pyproject.toml`](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html) should be undertaken as it is becoming the _de-facto_ method of configuring Python Packages.

* [x] Replace `versioneer` with `setuptools_scm`.
  * [x] `dynamic = ["version"]`
  * [x] `requires = ["setuptools >= 45", "setuptools_scm[toml]>=6.2", "wheel"]`
* [x] Update `docs/conf.py` to use `setuptools_scm`
* [x] Copy `setup.cfg` settings to `pyproject.toml`.
* [x] Add `check_toml` hook to `.pre-commit-config.yaml`.
* [x] Add [`Flake8-pyproject`](https://pypi.org/project/Flake8-pyproject/) as...
  * [x] `additional_dependencies` for `flake8` pre-commit hook.
  * [x] `project.optional.dependencies > dev` in `pyproject.toml`.
* [x] Configure [`ruff`](https://github.com/charliermarsh/ruff) in `pyproject.toml`.
* [x] Remove `setup.cfg` and `setup.py`.

ns-rse:
Closes #186

Switches to using [`pyproject.toml`](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html) alone for package configuration instead of a combination of `setup.cfg`/`setup.py`/`pyproject.toml` following [PEP621](https://peps.python.org/pep-0621/).

The following changes have been made...

* Switch from using [`versioneer`](https://github.com/claritychallenge/clarity/issues/222) to [`setuptools_csm`](https://github.com/pypa/setuptools_scm) to dynamically determine the package version.
* Introduces the need for [`Flake8-pyproject`](https://github.com/john-hen/Flake8-pyproject) as a dependency as flake8 is yet to switch to using this file for configuration (and is unlikely to do so, see links to relevant flake8 threads in [project description](https://github.com/john-hen/Flake8-pyproject)).
* De-duplicates optional dependencies between dev and tests.
* Removes pylint_pytest plugin from .pylintc as it cause pre-commit pylint hook to fail^[1].

Following #208 and de-duplication of files (thanks @jonbarker68 :+1: ) this package now builds locally and should do so as part of the PyPI workflow that publishes packages to PyPI and is triggered when a commit on `main` is tagged with `v#.#.#` (:crossed_fingers: ).

^[1] The `pylint` hook for `pre-commit` uses a virtual environment from the `pylint` repository which doesn't include `pylint_pytest` package and so the line in `.pylintrc` which loaded this plugin (`load-plugins=pytest_pylint`) always fails. The second `local` pipeline (`id: pylint_warnings`) _does_ load and include the `pytest_pylint` plugin because it is a `repo: local` and so runs on the users machine using their environment (this particular test is also therefore disabled from running on `pre-commit.ci`).


clarity/_version.py
```diff
@@ -1,696 +1,4 @@
...
-
-    return {
-        "version": "0+unknown",
-        "full-revisionid": None,
-        "dirty": None,
-        "error": "unable to compute version",
-        "date": None,
-    }
+# file generated by setuptools_scm
+# don't change, don't track in version control
```
jonbarker68, line 2: What does the 'don't track in version control' comment mean here?

pyproject.toml
```diff
@@ -30,21 +160,57 @@ exclude = '''
...
-[tool.coverage.run]
-omit = [
-    "clarity/_version.py"
 ]
-
-
-[tool.ruff]
-line-length = 88
-# pyflakes, pycodestyle, isort
-select = ["F", "E", "W", "B", "I001"]
```
jonbarker68, line 49: Have these ruff selections gone? I think by default ruff only applies F and E but I found W (codestyle-warnings) , B (flake8-bugbear) and I001 (unsorted-imports) useful too.

.pylintrc
```diff
@@ -45,7 +46,7 @@ limit-inference-results=100
 
 # List of plugins (as comma separated values of python module names) to load,
 # usually to register additional checkers.
-load-plugins=pylint_pytest
+load-plugins=
```
jonbarker68, line 49: Yes, I saw this was causing problems. I can't remember what it was needed for. I think pylint gets confused by fixtures without it.

jonbarker68 (changes requested):
Thanks! This is looking like it will be much simpler.  A few comments but possibly things I'm not understanding.

(Saw that the tests are weirdly failing though. Is that expected?)

.pylintrc
jonbarker68: Ah, just saw your explanation in the intro. That makes sense now.

pyproject.toml
ns-rse: Sorry, missed copying those over (twice, once when doing it and once when checking). Reinstated.

clarity/_version.py
ns-rse: Ah I should have dded this to `.gitignore` and excluded this from the git repository.

Its because the file is repeatedly auto-generated on commits.

ns-rse:
Weird that the tests are failing because we have explicitly included `importlib-metadata` as a dependency so it _should_ be installed. I'd created a new virtual environment to check the package built ok locally (it did) but didn't run the tests, having just done so these errors don't crop up.

Excuse the long post, can safely skip **Logs** and jump straight to **Summary** at the end...

## Logs
Checking the [logs](https://github.com/claritychallenge/clarity/actions/runs/4533505407/jobs/7986407831?pr=257)...

```
...
2023-03-27T15:07:24.7031489Z Collecting importlib-metadata
2023-03-27T15:07:24.7061535Z   Using cached importlib_metadata-6.1.0-py3-none-any.whl (21 kB)
...
2023-03-27T15:08:17.9903974Z Installing collected packages: pytz, msgpack, appdirs, antlr4-python3-runtime, zipp, wheel, urllib3, typing-extensions, tqdm, threadpoolctl, soupsieve, six, PyYAML, PySocks, pyparsing, pycparser, pillow, packaging, nvidia-cuda-nvrtc-cu11, numpy, llvmlite, lazy-loader, joblib, idna, future, fonttools, filelock, decorator, cycler, cloudpickle, charset-normalizer, certifi, audioread, submitit, soxr, scipy, requests, python-dateutil, omegaconf, nvidia-cuda-runtime-cu11, nvidia-cublas-cu11, kiwisolver, importlib-resources, importlib-metadata, cffi, beautifulsoup4, SoundFile, scikit-learn, pystoi, pyloudnorm, pooch, pandas, nvidia-cudnn-cu11, numba, matplotlib, hydra-core, torch, resampy, librosa, hydra-submitit-launcher, gdown, torchaudio, pyclarity
...
2023-03-27T15:09:19.2482225Z Successfully installed PySocks-1.7.1 PyYAML-6.0 SoundFile-0.12.1 antlr4-python3-runtime-4.8 appdirs-1.4.4 audioread-3.0.0 beautifulsoup4-4.12.0 certifi-2022.12.7 cffi-1.15.1 charset-normalizer-3.1.0 cloudpickle-2.2.1 cycler-0.11.0 decorator-5.1.1 filelock-3.10.7 fonttools-4.38.0 future-0.18.3 gdown-4.7.1 hydra-core-1.1.1 hydra-submitit-launcher-1.2.0 idna-3.4 importlib-metadata-6.1.0 importlib-resources-5.12.0 joblib-1.2.0 kiwisolver-1.4.4 lazy-loader-0.2 librosa-0.10.0.post2 llvmlite-0.39.1 matplotlib-3.5.3 msgpack-1.0.5 numba-0.56.4 numpy-1.21.6 nvidia-cublas-cu11-11.10.3.66 nvidia-cuda-nvrtc-cu11-11.7.99 nvidia-cuda-runtime-cu11-11.7.99 nvidia-cudnn-cu11-8.5.0.96 omegaconf-2.1.2 packaging-23.0 pandas-1.3.5 pillow-9.4.0 pooch-1.6.0 pyclarity-0.1.dev1+g7b27036.d20230327 pycparser-2.21 pyloudnorm-0.1.1 pyparsing-3.0.9 pystoi-0.3.3 python-dateutil-2.8.2 pytz-2023.2 requests-2.28.2 resampy-0.4.2 scikit-learn-1.0.2 scipy-1.7.3 six-1.16.0 soupsieve-2.4 soxr-0.3.4 submitit-1.4.5 threadpoolctl-3.1.0 torch-1.13.1 torchaudio-0.13.1 tqdm-4.65.0 typing-extensions-4.5.0 urllib3-1.26.15 wheel-0.40.0 zipp-3.15.0
...
2023-03-27T15:09:24.8620922Z Requirement already satisfied: importlib-metadata in /opt/hostedtoolcache/Python/3.7.16/x64/lib/python3.7/site-packages (from pyclarity==0.1.dev1+g7b27036.d20230327) (6.1.0)
...
2023-03-27T15:09:27.7218926Z Collecting importlib-metadata
2023-03-27T15:09:27.7310882Z   Downloading importlib_metadata-4.2.0-py3-none-any.whl (16 kB)
...
2023-03-27T15:09:34.2076044Z   Attempting uninstall: importlib-metadata
2023-03-27T15:09:34.2088215Z     Found existing installation: importlib-metadata 6.1.0
2023-03-27T15:09:34.2138677Z     Uninstalling importlib-metadata-6.1.0:
2023-03-27T15:09:34.2147447Z       Successfully uninstalled importlib-metadata-6.1.0
...
2023-03-27T15:09:36.4110640Z Successfully installed Flake8-pyproject-1.2.3 Jinja2-3.1.2 MarkupSafe-2.1.2 TOMLi-2.0.1 attrs-22.2.0 cfgv-3.3.1 coverage-7.2.2 distlib-0.3.6 exceptiongroup-1.1.1 flake8-5.0.4 flake8-print-5.0.0 identify-2.5.22 importlib-metadata-4.2.0 iniconfig-2.0.0 isort-5.11.5 mccabe-0.7.0 mypy-1.1.1 mypy-extensions-1.0.0 nodeenv-1.7.0 pathspec-0.11.1 platformdirs-2.6.2 pluggy-1.0.0 pre-commit-2.21.0 py-1.11.0 py-cpuinfo-9.0.0 pyclarity-0.1.dev1+g7b27036.d20230327 pycodestyle-2.9.1 pyflakes-2.5.0 pytest-7.2.2 pytest-cov-4.0.0 pytest-mpl-0.16.1 pytest-regtest-1.5.0 typed-ast-1.5.4 virtualenv-20.16.2 yamllint-1.30.0
```

## Summary

So initially `importlib-metadata-6.1.0` is installed, its then uninstalled in favour of `importlib-metadata-4.2.0` which shouldn't make a difference but I thought I'd investigate this.

I thought this might be down to multiple `pip install -e .` in the `Install package, including extras` step of the `run_tests.yml` as `pip` may not resolve dependencies correctly so set up a new virtualenv and went through each step listed with the `--no-cache-dir` to avoid using cache from my machine and emulating the situation on the runner. After each step I checked `pip show importlib-metadata` which version has been installed. Consistently it is `6.1.0`.

Just to be sure though I've combined the `pip install` stages to a single `pip install -e .[tests,docs,dev]` (arguably we could safely remove `docs` from this workflow as its used in the `gh-pages.yml` workflow).

Tests have passed on Python 3.8, 3.9 and 3.10 but failed on 3.7 (3.8 was cancelled originally but completed on a re-run before 3.7 failed and it was cancelled). There is something about 3.7 then that isn't adding up. I'll investigate further with a Conda environment but...

[3.7 is due to reach end of life 2023-06-27](https://devguide.python.org/versions/) and given #219 to support Python 3.11 maybe it would be prudent to instead bump the minimum supported version of Python 3.8.

This would involve...

1. Set `requires-python = ">=3.8"` in `pyproject.toml`.
2. Remove `3.7` from the test matrix `python-version: ['3.8', '3.9', '3.10']`

Perhaps this could be done under #219 if agreed upon?

ns-rse:
Ok I've replicated the error under Python 3.7, `importlib-metadata-4.2.0` was installed, explicitly upgrading to the most recent version `pip install --upgrade importlib-metadata` doesn't fix it either.

A solution (found [here](https://github.com/newville/asteval/pull/94/files)) is to use the following in `clarity/__init__.py`...

```
try:
    # python >=3.8
    from importlib.metadata import version, PackageNotFoundError
except ImportError:
    # python <3.8
    # importlib.metadata not available for python 3.7
    from importlib_metadata import version, PackageNotFoundError
```

But it depends on whether we want to continue supporting Python 3.7 or not given its imminent end-of-life.

jonbarker68:
I'm very happy to drop support for 3.7 which as you say is at end of life and given we should be able to start supporting 3.11 soon.

ns-rse:
> I'm very happy to drop support for 3.7 which as you say is at end of life and given we should be able to start supporting 3.11 soon.

Great, thanks for the confirmation, I've made the necessary changes in this Pull Request.

ns-rse:
One thing I forgot to mention when opening this PR is that I've added [`pytest-xdist`](https://pypi.org/project/pytest-xdist/) to the optional `test` dependencies. The fix @jonbarker68 merged the other day to correctly handle `random_matrix()` in tests (#254) means the order of tests now no longer matters and so the test suite can be run in parallel using this Pytest extension, if you've a 16 core machine you can therefore run...

```
pytest -n 12
```

Should help speed up the feedback loop locally.

jonbarker68 (approved):
Thanks. 

The pytest-xdist is a great addition. Should speed things up for some of us.

Issue #86, opened 2022-10-13
Unit tests for predictor module


ns-rse:
Develop unit tests for the data module (see #80 for overview).

Specific modules within `predictor` are...

- [ ] `predictor/torch_msbg.py`
- [ ] `predictor/torch_stoi.py`

jonbarker68:
torch_msbg contains functions that were already in evaluator/msbg and which appear to have been independently translated from the same MATLAB source. 

- generate_key_percent
- mearure_rms
- makesmearmat3
- audfilt

Some are not being used (replaced with torch implementation), e.g. measure_rms. so can delete. Others, e.. makesmearmat3 need to be deleted and replaced with calls to equivalents in evaluator/msbg -- need to also check that they are equivalent.

PR #259, opened 2023-03-28
86 unit tests for predictor module

jonbarker68:
Test code for the predictor modules. Will handle issue #86


jonbarker68:
@ns-rse - this PR should complete issue #86. It's mostly just tests of torch code but I have been able to remove a bit of code duplication along the way. The torch_msbg module was replicating some non-torch functions that were already in the evaluator/msbg modules.  Looks like merging this will bring test coverage up to around 90%.  👍

tests/predictor/test_torch_stoi.py
```diff
@@ -0,0 +1,178 @@
...
+from torch.testing import assert_close
+
+from clarity.predictor.torch_stoi import (
+    NegSTOILoss,
+    masked_mean,
+    masked_norm,
+    meanvar_norm,
+)
+
+# pylint: disable=redefined-outer-name,unused-argument  # pytest fixtures
```
ns-rse, line 15: These `pylint` errors are caused by having `pytest.fixtures` defined within the file as noted.

They can be avoided by placing fixtures in `tests/conftest.py` if desired.

ns-rse (approved):
Great to remove duplication and tests all look fine (not that I know what expected values should be!).

Minor suggestion on how to resolve the `pylint` error without disabling but great work blasting through all these testes @jonbarker68 really impressive to have such high test coverage.

tests/predictor/test_torch_stoi.py
jonbarker68: Thanks.  That's useful to know -- didn't realise conftests could be defined for each directory. Only problem with that is that it still defines the fixtures at the directory level which means that in some cases imports get pulled in for tests that don't need them. The real problem here I think is that pylint-pytest wasn't installed in my environment -- pylint-pytest is designed to fix some of these pytest-related false positives. I've now managed to check in ok without needing these disables.

ns-rse:
Does this still pass the `pre-commit` hook `pylint` even though `pylint-pytest` was removed as a module to be loaded as its not part of the virtual environment?

PR #261, opened 2023-03-28
260-increase-pyupgrade-version

jonbarker68:
Changing pyupgrade argument to --py38-plus in commit hook and rerunning.


jonbarker68:
@groadabike, could you approve this one please. It's just a trivial update that drops code that we don't need now that we are no longer supporting Python 3.7

ns-rse (approved):
Sorry I missed this out.

PR #264, opened 2023-03-29
Align black and ruff target-version to py38 with rest of package

ns-rse:
I missed these out too and only just noticed.


jonbarker68 (approved):
Thanks

PR #269, opened 2023-03-31
252 deprecation warnings in tests

### LINKED ISSUE #252: Deprecation warnings in tests
Several deprecation warnings appear when running the test suite.

 These should be dealt with

- [ ] NumbaPendingDeprecationWarning:  'reflected list' in HAO_tools_cec2.py
- [ ] DeprecationWarning: `np.int` is a deprecated alias for buildin 'int' in torch_msbg.py
- [ ] FutureWarning: `rcond` parameter - in torch_msbg.py 

The torch_msbg.py issues may be fixed in issue #86 dealing with torch_msbg tests which will include some torch_msbg refactoring.

jonbarker68:
Minor edits to fix warnings being thrown when running pytest. 

Closes #252

### DIFF
--- a/tests/recipes/cpc2/test_evaluate_cpc2.py
+++ b/tests/recipes/cpc2/test_evaluate_cpc2.py
@@ -1,6 +1,7 @@
 """Tests for the CPC2 evaluation functions."""
 
 import math
+import warnings
 
 import numpy as np
 import pytest
@@ -30,6 +31,8 @@ def test_rmse_score_ok(x, y, expected):
 def test_rmse_score_error(x, y, expected):
     """Test the function rmse_score for invalid inputs"""
     with pytest.raises(expected):
+        np.seterr(all="ignore")
+        warnings.simplefilter("ignore")  # <--- suppress mean of empty slice warning
         result = rmse_score(np.array(x), np.array(y))
         if np.isnan(result):
             raise ValueError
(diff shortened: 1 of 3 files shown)


jonbarker68:
pytest was reporting some warnings that have now been dealt with.

- warnings from numpy in tests that were designed to throw exceptions. These have been disabled to avoid noise.
- warning about obsolete np.int32 type that have been fixed
- the audiogram code was relying on the mean of an empty array being 0. Fixed the logic to avoid this.

pytest will now run without producing any warningings

This closes issue #252

tests/recipes/cpc2/test_evaluate_cpc2.py
```diff
@@ -30,6 +31,8 @@ def test_rmse_score_ok(x, y, expected):
 def test_rmse_score_error(x, y, expected):
     """Test the function rmse_score for invalid inputs"""
     with pytest.raises(expected):
+        np.seterr(all="ignore")
+        warnings.simplefilter("ignore")  # <--- suppress mean of empty slice warning
```
groadabike, line 35: you can catch the warnings with the context manager. this will deactivate and reactivate the warnings
```
with warnings.catch_warnings():
   warnings.simplefilter("ignore")
```

groadabike (approved):
great to catch all the test warnings.
Just a comment using catch_warning context manager.
I am not sure how different is the behaviour compared to your version.

I assume that having the warnings simple filter within the pytest.raises context, deactivates the warnings only in that context

tests/recipes/cpc2/test_evaluate_cpc2.py
jonbarker68: Thanks, this looks better.

Issue #270, opened 2023-04-01
Tests for recipes


jonbarker68:
Test coverage for the recipes is very poor. 

In order to safely refactor the core library we will need recipe tests to ensure results remain reproducible. End-to-end regression tests with very small amounts of data will be sufficient.

- [ ] recipes/cad1/task1 (**partially complete**)
- [ ] recipes/cad1/task2 (**partially complete**)
- [x] recipes/cec1/baseline
- [x] recipes/cec1/data_preparation/prepare_cec1_data.py
- [x] recipes/cec1/e009_sheffield
- [x] recipes/cec2/baseline
- [x] recipes/cec2/data_preparation
- [x] recipes/cpc1/baseline
- [x] recipes/cpc1/e029_sheffield
- [x] recipes/cpc1/e032_sheffield
- [x] recipes/cpc2  (**partially complete**)
- [x] recipes/icassp_2023

## Baseline recipes vs contributed recipes

The priority in this issue is to ensure all the baseline recipes have good test coverage. For cpc1/e029 and cpc1/e032, only the data preparation scripts are being tested.

PR #271, opened 2023-04-03
270-tests-for-recipes

jonbarker68:
# Improved test coverage for the recipes

This PR is mostly new tests for recipes and associated test data. A few minor issues
with the recipes have been fixed along the way.

## Test and test data

- Effort has been made to make sure that the tests are quick to run.Tests that take more
than about 10 seconds are marked with slow and will be skipped by default. These can be
run with `pytest -m slow`. The entire suite of now over 400 tests runs in less than
3 mins on GH actions and in about 1 min locally on a macbook pro.

- The test data has been kept as small as possible with extensive use of symbolic links
to allow files to be shared. Note, test files have standard clarity names but can have
fake content. i.e. don't expect them to be the same as the real distributed data.

- There is a certain amount of redundancy in the test code. e.g. a lot of tests are
using a one-line function to mock out and silence tqdm. I consider refactoring some
of this but that can wait for now but in most places it would just make things less
explicit.

## Recipe coverage

- All the official baseline recipes have good coverage. i.e. ~90%, except for cad1 
task1 and task2 which are missing tests for the top-level functions. Can be added later.
- There is poor coverage for the 'contributed' cpc1 recipes. This is not a priority.

## Modifications to the recipes themselves

- Lack of tests meant that some of the recipes were broken and needed fixing. The 
common issue was where parameter names to functions in the library had changed. 
- The recipes are often using config dicts to pass parameters, e.g. **cfg.Ear, to
allow this to continue to work the parameter names in the config yaml files have
been updated where necessary.
- A significant BUG in cec1/e009_sheffield/train.py has been fixed. Here an f-string
f"{ear}_amp" was being used in path names but the 'f' had been missed. This will have
led to both ears sharing the same directory for storing models.

### DIFF
--- a/clarity/evaluator/haspi/haspi.py
+++ b/clarity/evaluator/haspi/haspi.py
@@ -165,15 +165,19 @@ def haspi_v2_be(  # pylint: disable=too-many-arguments
 
     # Adjust listener.audiogram_levels_l and _r to match the frequencies above
     hearing_loss_left = [
-        audiogram_left[i]
-        for i in range(len(audiogram_frequencies))
-        if audiogram_frequencies[i] in aud
+        loss
+        for (freq, loss) in zip(audiogram_frequencies, audiogram_left)
+        if freq in aud
     ]
+
     hearing_loss_right = [
-        audiogram_right[i]
-        for i in range(len(audiogram_frequencies))
-        if audiogram_frequencies[i] in aud
+        loss
+        for (freq, loss) in zip(audiogram_frequencies, audiogram_right)
+        if freq in aud
     ]
+
+    if len(hearing_loss_left) != len(aud) or len(hearing_loss_right) != len(aud):
+        raise ValueError("Audiogram does not have all measurements needed by HASPI.")
 
     score_left, _ = haspi_v2(
         reference_left,
--- a/tests/evaluator/haspi/test_haspi.py
+++ b/tests/evaluator/haspi/test_haspi.py
@@ -21,15 +21,29 @@ def test_haspi_v2() -> None:
     )
 
 
-def test_haspi_v2_better_ear() -> None:
+@pytest.mark.parametrize(
+    "hl_left, hl_right, freqs, expected_score",
+    [
+        (
+            np.array([25, 25, 25, 25, 40, 65]),
+            np.array([45, 45, 35, 45, 60, 65]),
+            np.array([250, 500, 1000, 2000, 4000, 6000]),
+            0.839975335323691,
+        ),
+        (
+            np.array([25, 25, 21, 25, 25, 40, 65]),
+            np.array([45, 45, 21, 35, 45, 60, 65]),
+            np.array([250, 500, 700, 1000, 2000, 4000, 6000]),  # <-- extra cf added
+            0.839975335323691,
+        ),
+    ],
+)
+def test_haspi_v2_better_ear(hl_left, hl_right, freqs, expected_score) -> None:
     """Test for hasqi_v2_better_ear index"""
 
-    np.random.seed(0)
     sample_rate = 16000
-    hl_left = np.array([25, 25, 25, 25, 40, 65])
-    hl_right = np.array([45, 45, 35, 45, 60, 65])
 
-    freqs = np.array([250, 500, 1000, 2000, 4000, 6000])
+    np.random.seed(0)
     ref_left = np.random.uniform(-1, 1, int(sample_rate * 0.5))  # i.e. 500 ms of audio
     ref_right = np.random.uniform(-1, 1, int(sample_rate * 0.5))
     proc_left = np.random.uniform(-1, 1, int(sample_rate * 0.5))  # i.e. 500 ms of audio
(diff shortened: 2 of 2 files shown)


groadabike:
> Adding tests for all the recipes.
> 
> The recipes are moved into the clarity/ package directory. There didn't seem to be a strong motivation for having 'recipes' outside of the clarity package. Placing everything under 'clarity' makes things a lot easier to manage and simplifies the packaging.

Asteroid and SpeechBrain have the `Recipes` outside the packages. 
I presume this is to separate the package functionalities from the examples of using the package.

Tbh, I prefer having the recipes outside because we expect people to adapt the recipes to their needs, having the recipes inside `clarity` may feel more like messing with the package than modifying a recipe

jonbarker68:
> 

OK, that's a good point. And it's best to keep things as conventional as possible. We'll keep them outside. But we'll need to adjust some of the config files to make sure that they are included in the code coverage. I'll move them back up before completing this PR

jonbarker68:
@groadabike @ns-rse - this is a bit of an epic PR but it's mostly just test coverage for the recipe code that had previously been overlooked -- there are also a few minor fixes to the recipes themselves that have been made along the way as issues were discovered. I've added a short intro at the top to make it easier to follow. A lot of files are reported as changed but these are mostly additional small files under tests/test_data.  This deals with  most of issue #270 and should mean we are at much less risk of breaking recipes when refactoring.  😄

clarity/enhancer/gha/gha_interface.py
```diff
@@ -64,7 +64,7 @@ def create_configured_cfgfile(
             cfg_filename (str): cfg filename
         """
 
-        if self.fs != 44100:
+        if self.sample_frequency != 44100:
```
groadabike, line 67: This is a name we need to agree with. I always use `sample_rate`, I can see you use `sample_frequency`. I am happy with either.  I can re-name later mine with sample_frequency if you prefer.

clarity/evaluator/haspi/haspi.py
```diff
@@ -132,10 +132,10 @@ def haspi_v2_be(  # pylint: disable=too-many-arguments
     reference_right,
     processed_left,
     processed_right,
-    fs_signal,
+    sample_freq,
```
groadabike, line 135: here is another version of the name, `sample_freq`

clarity/evaluator/haspi/haspi.py
```diff
@@ -165,20 +165,29 @@ def haspi_v2_be(  # pylint: disable=too-many-arguments
 
     # Adjust listener.audiogram_levels_l and _r to match the frequencies above
     hearing_loss_left = [
-        audiogram_left[i] for i in range(len(audiogram_cfs)) if audiogram_cfs[i] in aud
+        audiogram_left[i]
+        for i in range(len(audiogram_frequencies))
+        if audiogram_frequencies[i] in aud
     ]
```
groadabike, line 171: I am wondering if the scores should be responsible for this validation. In fact, the only thing this is doing is just filtering the frequencies if the `audiogram_frequencies` are in `aud` but there is no validation that all the `hearing_loss_left` has an audiogram value for each frequency in `aud`

I would consider later making the user responsible for having an audiogram for each frequency and the score validates that all the `audiogram_frequencies` are included.

tests/recipes/cad1/task1/baseline/test_evaluate.py
```diff
@@ -182,3 +182,9 @@ def test_evaluate_song_listener(
             rel=pytest.rel_tolerance,
             abs=pytest.abs_tolerance,
         )
+
+
+@pytest.mark.skip(reason="Not implemented yet")
+def test_run_calculate_aq():
+    """test run_calculate_aq function"""
+    # run_calculate_aq()
```
groadabike, line 190: This seems to be  homework for me :)

tests/recipes/cad1/task2/baseline/test_audio_manager.py
```diff
@@ -32,3 +32,18 @@ def test_save_audios(tmp_path):
...
+
+
+@pytest.mark.skip(reason="Not implemented yet")
+def test_get_lufs_level():
+    """Test get_lufs_level method."""
+
+
+@pytest.mark.skip(reason="Not implemented yet")
+def test_scale_to_lufs():
+    """Test scale_to_lufs method."""
```
groadabike, line 49: More homework

jonbarker68: Yes, but perhaps not urgent and I doesn't necessarily need a test for each function. For some of the CEC, CPC recipes I felt it was sufficient to test the top level run function that then covers the low level functions. This will still pick up a breaking change after which it wouldn't be hard to isolate the function where it happens.

clarity/evaluator/haspi/haspi.py
jonbarker68: Yes. agree. For now, I will at least fix it so that it throws a ValueError if a needed frequency is missing. This will look different when audiogram objects are used anyway.

PR #274, opened 2023-04-04
272 tweak score for task1

groadabike:
Add a score for CAD1 - TASK 1
HAAQI + RMS


groadabike:
The PR includes:
1.  New score that combines HAAQI (from the music) and RMS (from the silence part of the processed signal).
2. Changes in Task 1 recipe to use the new score
3. Changes in Task 1 to scale processed signal to rms=1 before HA as haaqi documentation says

jonbarker68:
> @jonbarker68 Currently, the submission is massive because we are saving the output VDBO using 44100 Hz and Float64 These are the signals that go to HAAQI. Do you think is ok to save them in 16bit?

Yes, I think actually it's the more correct thing to do.  We may need to use 16bit for the listening tests (i.e. the USB headphones are unlikely to support floating point) and it is more consistent if HAAQI and listening tests are based on the same signals.  This introduces the need to avoid overflow (i.e. clipping) -- with floating point into to HAAQI the signals can have an arbitrarily large dynamic range -- with 16 bit they will clip if they get too large and this is something that competitors need to deal with.

jonbarker68 (commented):
Can we split this into two separate PRs - i.e. changes to the recipes/tests in one and changes to haspi and haaqi in another. We need to be very careful with changes to 'eb.py' as recipes for CEC1, CEC2, CPC1, CPC2 and CEC2_ICASSP all depend on it.  The recipes have not yet got sufficient testing in place -- this is being fixed by PR #270 but this is not yet in main.

recipes/cad1/task1/baseline/enhance.py
```diff
@@ -247,6 +261,25 @@ def process_stems_for_listener(
     return processed_stems
 
 
+def clip_and_save(signal: np.ndarray, filename: Path | str, config: DictConfig) -> None:
```
jonbarker68, line 264: Wouldn't it be better to just have a function to clip and return a result which is then saved. This seems to be merging two tasks. 

I would also avoid passing config dictionaries. Clearer and more explicit to just provide a soft_clip and sample_rate parameter.

recipes/cad1/task1/baseline/enhance.py
```diff
@@ -385,25 +422,18 @@ def enhance(config: DictConfig) -> None:
...
-        # Clip and save
-        if config.soft_clip:
-            enhanced = np.tanh(enhanced)
-        n_clipped = np.sum(np.abs(enhanced) > 1.0)
-        if n_clipped > 0:
-            logger.warning(f"Writing {filename}: {n_clipped} samples clipped")
-        np.clip(enhanced, -1.0, 1.0, out=enhanced)
-        signal_16 = (32768.0 * enhanced).astype(np.int16)
-        wavfile.write(filename, config.nalr.fs, signal_16)
+        clip_and_save(enhanced, filename, config)
```
jonbarker68, line 436: Yep, rewrite as 
```
clipped_signal = clip(
wavfile.write(... , clipped_signal)
```

tests/recipes/cad1/task1/test_evaluate.py
```diff
@@ -101,14 +100,14 @@ def test_make_song_listener_list():
...
             {
-                "left_drums": 0.149880148,
-                "right_drums": 0.143182857,
-                "left_bass": 0.140449345,
-                "right_bass": 0.181374074,
-                "left_other": 0.132401105,
-                "right_other": 0.164211137,
-                "left_vocals": 0.121260627,
-                "right_vocals": 0.126605279,
+                "left_drums": 0.140304064,
```
jonbarker68, line 103: Some of these scores have changed significantly. Was that expected?

jonbarker68 (changes requested):
Good stuff. A couple of minor points.

I was a bit worried about combining the eb.py work and CAD1 work in the same PR. But the eb.py changes all look good. If we are sure that the test coverage of eb is good and that the changes have made not difference to the eb behaviour then I think it is safe enough.

clarity/evaluator/haspi/eb.py
```diff
@@ -1,4 +1,5 @@
 """Module for HASPI, HASQI, HAAQI EBs"""
+# pylint: disable=import-error
```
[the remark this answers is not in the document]
groadabike: I run the `test_eb.py` after each change to ensure the same results.  
I don't know about the coverage.

tests/evaluator/haspi/test_eb.py
```diff
@@ -293,7 +292,7 @@ def test_env_compress_basilar_membrane():
         fsamp=24000,
         level1=140,
         small=1e-30,
-        threshold_high=120,
+        threshold_high=100,
```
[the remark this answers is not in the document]
groadabike: Because this parameter had no impact in the test as the `threshold_high` was then hardcoded to 100 in the `env_compress_basilar_membrane` function. I change the test to 100 to preserve the previous results

clarity/evaluator/haspi/eb.py
```diff
@@ -876,9 +875,6 @@ def env_compress_basilar_membrane(
     Two-tone suppression added 22 August 2008.
     Translated from MATLAB to Python by Zuzanna Podwinska, March 2022.
     """
-    # Initialize the compression parameters
-    threshold_high = 100
-
```
groadabike, line 881: Because I remove this hard-coded value of `threshold_high`, I had to change the test to 100

clarity/utils/signal_processing.py
```diff
@@ -1,18 +1,20 @@
+"""Signal processing utilities."""
```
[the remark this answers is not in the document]
groadabike: No, i didn't change the function in any way, just reorder

recipes/cad1/task1/baseline/enhance.py
groadabike: I put both tasks together as all audios are clipped and saved. But, this can be split into 2 functions and perhaps move them to `clarity/utils`

tests/recipes/cad1/task1/test_evaluate.py
groadabike: Yes, because we add the RMS factor to the score.
Do, the score now is a combination of HAAQI and RMS of the silence parts.

Thai change of the score was part of the discussion with Trevor

groadabike:
> > @jonbarker68 Currently, the submission is massive because we are saving the output VDBO using 44100 Hz and Float64 These are the signals that go to HAAQI. Do you think is ok to save them in 16bit?
> 
> Yes, I think actually it's the more correct thing to do. We may need to use 16bit for the listening tests (i.e. the USB headphones are unlikely to support floating point) and it is more consistent if HAAQI and listening tests are based on the same signals. This introduces the need to avoid overflow (i.e. clipping) -- with floating point into to HAAQI the signals can have an arbitrarily large dynamic range -- with 16 bit they will clip if they get too large and this is something that competitors need to deal with.

Great!!
That makes sense. And, this will reduce the size of the submission in half or more.

jonbarker68 (approved):
Looks good.  But would change clip_and_save to take separate parameters rather than a configuration dictionary.

jonbarker68:
@groadabike  feel free to merge this when it's ready.

PR #276, opened 2023-04-10
219 support for python 311

jonbarker68:
# Support for Python 2.11

Python 2.11 has required some key packages to need major version upgrades.

- torch >= 2.0
- torchaudio >= 2.0
- numba >= 0.57.0rc

The upgrade to torchaudio has required a few further fixes

- `stft` now returns a complex tensor meaning `torch_msbg` needed some modifications.
- `transforms.Resample` now has different names for its windowing parameters

## Other changes

- For some reason, the cpu-denpendence of the torch_msbg_stoi regression test has
re-emerged so that commented out `skip-if` logic has been re-instated.
- `dataclass` now prohibits direct initialisation with mutable lists and so any
small change to `audiogram.py` has been needed.

All tests are now passing and this could be merged into main. It will however force
any users to upgrade their torch and torchaudio packages.


jonbarker68:
Lots of torch-based regression tests are failing. I believe this is because the random numbers have changed. Either the seeding needs fixing or it may be that even with the correct seeding there is no way to maintain reproducibility across torch versions. If that's the case we may need to i) temporarily change tests to run off saved random data , ii) used the save data to make sure tests pass and then iii) revert to seeding and change the expected outputs.

groadabike:
> The following torch-based regression tests are failing.
> 
> ```
> tests/predictor/test_torch_stoi.py ....FFFF.......
> tests/regression/test_engine_losses.py ..FF
> tests/regression/test_mc_conv_tasnet.py .F
> tests/regression/test_predictors.py Fs.
> ```
> 
> I believe this may be because the random numbers have changed (this hypothesis needs checking...). In which case either the seeding needs fixing or it may be that even with the correct seeding there is no way to maintain reproducibility across major torch versions. If that's the case we may need to i) temporarily change tests to run off saved random data , ii) used the save data to make sure tests pass and then iii) revert to seeding and change the expected outputs.

@jonbarker68 
Indeed they don't guarantee reproducibility across Pytorch versions https://pytorch.org/docs/stable/notes/randomness.html

In terms of the random seed, Pytorch 1.13 and 2.0 seem to be returning the same values.
Pytorch 1.13
image

Pytorch 2.0
image

jonbarker68:
> Perhaps, one solution is to set the `deterministic_algorithm` to `TRUE` if is not already https://pytorch.org/docs/stable/notes/randomness.html#avoiding-nondeterministic-algorithms

@groadabike Yes, on further investigation the issue does not appear to be the random numbers. I expect it's more likely to be numeric issues. e.g. we are getting small but significant differences in low STOI loss scores e.g., 0.0003 vs 0.0140. But we are comparing random signals that have no correlation.  I expect these results are very sensitive to precision errors. I'll need to do a bit more digging.

jonbarker68:
@groadabike - the main problem turned out to be in `torchaudio.transforms.Resample` where they has been a change to the valid values for the 'resampling_method' parameter: before it was called "sinc_interpolation" now it is "sinc_interp_hann" ... but weirdly it didn't complain that the old value was no longer valid it just went and accepted and used some different method. That feels like a bug in the torchaudio code TBH.   Now just one test failing.

jonbarker68:
All tests now passing but pytest has started to run very slowly on GitHub actions.  Up to 10 minutes vs <30 seconds on my mac. Was only around 4 minutes before so not sure what has changed.

jonbarker68:
This is now ready to be merged into main but probably best to wait until #270 is merged so that the recipe test coverage is increased. We can then merge main into this branch and check that all is still OK before final merge into main.

jonbarker68:
This PR could now be merged but doing so will require users to upgrade their torch version.

clarity/_version.py
ns-rse, line 1: I should have removed this file from the repository as per the instructions at the top.

Will do so under #284

PR #283, opened 2023-04-11
282 fixing all non strict mypy issues

jonbarker68:
# Refactoring and fixes to handle all issues raised by MyPy

After this PR, `mypy` should run without finding any issues

```bash
✦ 🕙 12:34:18 ✖  mypy .
Success: no issues found in 219 source files
```

Summary of what's been done:

- Using monkeytype and manual checking to complete type annotation on key parts of the 
code base
- Fixing type issues arising from parameters defaulting to integer literals when they
should have been floats
- Refactoring some particularly nasty bits of the codebase - in particular the code
for computing camfit gaintables.
- Some minor edits and improved variable naming 

Functions expecting arrays have been typed as `np.ndarray`. These functions could have
been typed with `ArrayLike` <https://numpy.org/devdocs/reference/typing.html>, 
which would allow them to accept lists as well as arrays.
But this requires a lot of fussy code like `x = x_input.asarray()` at the start of 
every function and if it is not done consistently it can be hard for users to know 
which functions accpet lists and which don't. Better if is the user's responsibility 
to convert lists to arrays before calling.


clarity/enhancer/compressor.py
```diff
@@ -1,41 +1,42 @@
...
+
 
 class Compressor:
     def __init__(
         self,
-        fs: int = 44100,
-        attack: int = 5,
-        release: int = 20,
-        threshold: float = 1,
+        fs: float = 44100.0,
```
groadabike, line 14: Why was `fs` changed to float? 
Perhaps, we could change this parameter to `sample_rate` like NAL-R which usually are used together.

groadabike (approved):
All look good to me.
Just wondering why change the `sample_rate` to float.

Also, here we could change the `fs` param in Compressor to `sample_rate`

clarity/enhancer/compressor.py
jonbarker68: It's more 'correct' I think and less 'opinionated':

- Samples per second can be an arbitrary number. It just happens that standard audio formats quote sample rates as integers 16000, 44100 etc.  If we were to be passing a downsampled signal to a function we may very well have a fractional sampling rate.

- Float is more permissive: i.e. float will accept floats or ints; but int will complain if it receives a float -- even a float that happens to be roundable to an int such as 16000.0.  So my approach has been to start off with quite permissive types -- we can always choose to restrict further later. 


```
def fn_int(x_int: int):
    print(x_int)

def fn_float(x_float: float):
    print(x_float)

fn_int(1)  # OK, obviously
fn_float(1.0)  # OK, obviously
fn_int(1.0)  # <-- not OK
fn_float(1)  # <-- but also ok as float can accepts ints or floats
```

- ints can become floats in unexpected ways: e.g. ints turn into floats on division -- even if they are even -- unless you are using integer division '//' -- which is not used consistently in the repo, e.g. in torch_msbg there is "nyquist = sr / 2"  

```
In [6]: type(16000/2)
Out[6]: float
```

PR #290, opened 2023-04-15
277 unifying audiogram code

jonbarker68:
Working towards consistent use of a unified audiogram class


jonbarker68:
@groadabike this is now mostly complete.  I think it's a big improvement on what we had before but would appreciate a review.

It's not quite ready to merge yet: The audiogram is currently in enhancer/msbg/audiogram.py but needs moving to somewhere higher up as its used in camfit, msbg, haspi etc.  utils/ is one option but there is a danger then that utils would just end up with a lot of random stuff.

clarity/utils/audiogram.py
```diff
@@ -0,0 +1,170 @@
...
+            new_frequencies (ndarray): The new frequencies to resample to
+
+        Returns:
+            Audiogram: New audiogram with resampled frequencies
+
+        """
+        return Audiogram(
+            levels=logx_interp(self.frequencies, self.levels, new_frequencies),
+            frequencies=new_frequencies,
+        )
```
groadabike, line 139: Perhaps, we could validate first that `self.frequency` and `new_frequency` are different before computing the `logx_interp`.
If they are equal, return the `self.Audiogram`

groadabike (changes requested):
This looks much better now with a single `Audiogram` class.
Just 2 comments:
* Maybe we could validate that the sample frequencies are different before doing the resample.
* I agree with you the `utils` doesn't sound like the best place. Audiograms sound like a `core` part of any of our challenges, what do you think to create a `core` directory and save it there? I am not sure what else would be core though.

tests/evaluator/haspi/test_haspi.py
```diff
@@ -90,15 +94,14 @@ def test_haspi_v2_better_ear_error(hl_left, hl_right, freqs) -> None:
...
+        reference_left=ref_left,
+        reference_right=ref_right,
+        processed_left=proc_left + ref_left,
+        processed_right=proc_right,
+        sample_rate=sample_rate,
+        audiogram_left=Audiogram(levels=hl_left, frequencies=freqs),
+        audiogram_right=Audiogram(levels=hl_right, frequencies=freqs),
+        level=100,
+    )
+    assert score == pytest.approx(expected_score)
```
groadabike, line 107: Here are mising the `rel=pytest.rel_tolerance, abs=pytest.abs_tolerance` params in pytest.approx

tests/evaluator/msbg/test_audiogram.py
```diff
@@ -43,3 +43,63 @@ def test_standard_audiograms_are_correct_severity():
...
+def test_resample(requested_frequencies, expected_levels):
+    """test that a subset of cfs can be chosen"""
+    levels = np.array([1, 2, 3, 4, 5, 6])
+    frequencies = np.array([250, 500, 1000, 2000, 4000, 6000])
+    audiogram = Audiogram(levels=levels, frequencies=frequencies)
+
+    # Include cfs that don't exist in the requested subset
+    subset_audiogram = audiogram.resample(requested_frequencies)
+    assert np.allclose(subset_audiogram.frequencies, requested_frequencies)
+    assert np.allclose(subset_audiogram.levels, expected_levels)
```
groadabike, line 105: I thought you prefer the `pytest.approx` approach to `np` functions for validate test.

tests/predictor/test_torch_msbg.py
groadabike, line 1: In this module, there are several `pytest.approx` without the `rel=pytest.rel_tolerance, abs=pytest.abs_tolerance`

tests/recipes/cad1/task1/baseline/test_enhance_task1.py
groadabike, line 1: In almost all of my tests in CAD1, I used the `np.tesing` functionalities, I may need to change them later to use `pytest.approx`

groadabike (changes requested):

tests/evaluator/msbg/test_audiogram.py
jonbarker68: Yes. Good point. I'll fix this.

tests/predictor/test_torch_msbg.py
jonbarker68: Yes. We should remove them everywhere as they are the default values. See point above.

clarity/utils/audiogram.py
jonbarker68: Thanks. I'll reconsider this.

For that one case alone, I'm not sure the efficiency saving is worth the complication. The current function returns a new audiogram so you would still need to make a copy or the behaviour would be unpredictable so you would end up with something like,

```
 if np.all(new_frequencies == self.frequencies):
            return copy.deepcopy(self)
```

But elsewhere in the code though, we had a method for making an audiogram from a subset of existing audiogram frequencies which is now being done using the resample method -- but this doesn't validate that the subset existed. So perhaps the resample could have a 'validate' flag the is defaults to False but if True will only work if the new_frequencies were a subset ... so that you can guarantee that the new audiogram is made of genuine HL measurements not interpolated.

tests/predictor/test_torch_msbg.py
jonbarker68: I was wrong above so I have now added them everywhere insteead.

jonbarker68:
@groadabike - I've addressed the points raised in your review -- except package is still in utils. I think we'll need to leave it there for now. Happy to merge this if you are.

groadabike (approved):
All look good to me

PR #294, opened 2023-04-17
293-refactor-signal-read-and-write

jonbarker68:
# Refactoring of audio input and output

Similar functions for reading and writing audio files were previously defined at several points in the library. These have all been replaced by a single pair of functions in utils/file_io that are better documented and provide consistent behaviour.

### DIFF
--- a/clarity/evaluator/msbg/msbg_utils.py
+++ b/clarity/evaluator/msbg/msbg_utils.py
@@ -10,7 +10,6 @@ from typing import Final, TypedDict
 import numpy as np
 import scipy
 import scipy.signal
-import soundfile
 from numpy import float64, ndarray
 from soundfile import SoundFile
 
@@ -566,45 +565,3 @@ def read_signal(
         )
 
     return signal
-
-
-def write_signal(
-    filename: str | Path,
-    signal: ndarray,
-    sample_rate: float,
-    floating_point: bool = True,
-    strict: bool = False,
-) -> None:
-    """Write a signal as fixed or floating point wav file.
-
-    NB: setting 'strict' to True will raise error on overflow. This would be
-    a more natural default but it would break existing code that did not
-    check for overflow.
-
-    Args:
-        filename (str|Path): name of file in to write to.
-        signal (ndarray): signal to write.
-        sample_rate (float): sampling frequency.
-        floating_point (bool): write as floating point else an ints (default: True).
-        strict (bool): raise error if signal out of range for int16 (default: False).
-    """
-
-    if sample_rate != MSBG_FS:
-        logging.warning(
-            f"Sampling rate mismatch: {filename} with "
-            f"sample frequency = {sample_rate}."
-        )
-        # raise ValueError("Sampling rate mismatch")
-
-    if floating_point is False:
-        subtype = "PCM_16"
-        # Signal is float and we want to convert to int16
-        # *NB* Not  *= in next line as we need to make a copy
-        signal = signal * 32768
-        if strict and (np.max(signal) > 32767 or np.min(signal) < -32768):
-            raise ValueError("Signal out of range -1.0 to 1.0")
-        signal = signal.astype(np.dtype("int16"))
-    else:
-        subtype = "FLOAT"
-
-    soundfile.write(filename, signal, sample_rate, subtype=subtype)
--- a/clarity/evaluator/msbg/msbg_utils.py
+++ b/clarity/evaluator/msbg/msbg_utils.py
@@ -524,7 +524,7 @@ def pad(signal: ndarray, length: int) -> ndarray:
     )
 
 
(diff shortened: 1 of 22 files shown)


jonbarker68:
@groadabike - I had initially wanted to store multichannel signals as shape  [n_channel, n_samples] as this makes it easier to iterate over channels i.e.  'for channel in signal:'  but soundfile (which is widely used) uses the opposite convention, so I think transpose on read and write would confuse people.  Iterating over channels is still easy enough with 'for channel in signal.T:'

tests/recipes/cec1/baseline/test_evaluate.py
```diff
@@ -31,15 +31,21 @@ def not_tqdm(iterable):
 def truncated_read_signal(
```
ns-rse, line 31: Would it be reasonable to make `utils.file_io.read_signal()` have an option of how much of a signal to return (or write for `utils.file_io.write_signal()`)?

clarity/utils/file_io.py
```diff
@@ -14,3 +25,131 @@ def write_jsonl(filename: str, records: list) -> None:
...
+#   the file is resample if necessary or can be set to throw an error. Again,
+#   this is more convenient for most purposes.
+
+
+def write_signal(
+    filename: str | Path,
+    signal: ndarray,
+    sample_rate: float,
+    floating_point: bool = True,
+    strict: bool = False,
```
ns-rse, line 51: I realise it was only a test in which it was used but was wondering if it might be of general use to have an option in `write_signal()` and `read_signal()` to specify how much of a sample to read/write?

clarity/evaluator/msbg/msbg_utils.py
```diff
@@ -525,7 +524,7 @@ def pad(signal: ndarray, length: int) -> ndarray:
     )
 
 
-def read_signal(
+def read_signal_x(
```
ns-rse, line 527: Could this function not be removed too? Code looks very similar, at the minimum could have the file read by `utils.file_io.read_signal()` and then resample `if wave_file.samplerate != MSBG_FS:`.

clarity/utils/file_io.py
```diff
@@ -14,3 +25,131 @@ def write_jsonl(filename: str, records: list) -> None:
...
+    if not offset_is_samples:  # Default behaviour
+        offset = np.rint(offset * wave_file.samplerate).astype(int)
+
+    if offset != 0:
+        wave_file.seek(offset)
+
+    signal = wave_file.read(frames=n_samples)
+
+    if sample_rate not in (0, wave_file.samplerate):
+        if allow_resample:
```
ns-rse, line 145: I see in `msbg_utils.read_signal_x()` that resampling is done and wonder if having a separate function for resampling might be more useful rather than having it embedded within the `read_signal()` function.

It would allow for a specific unit-test to be written and resampling to be performed anywhere within the code base.

More appropriate as separate issue though I think (if its a useful thing to do).

ns-rse (approved):
Great simplification and reduction of duplicate code.

Some minor suggestions in line which might further abstract functionality and reduce one remaining area of duplication. Probably for a separate issue though.

clarity/evaluator/msbg/msbg_utils.py
jonbarker68: Yes! Thanks. The rename to read_signal_x was just a step I made during porting to quickly see where it was being used :-). I then forgot to delete the function.

tests/recipes/cec1/baseline/test_evaluate.py
jonbarker68: Thanks. Actually, read_signal already has that. So this should be rewritten to use it, i.e like

```
signal = read_signal(
        filename,
        offset=offset,
        n_samples=n_samples,
        n_channels=n_channels,
        offset_is_samples=offset_is_samples,
        n_samples=44100,
    )
```

clarity/utils/file_io.py
jonbarker68: Yep. It's in read_signal().  I think it's less well motivated for write signal as it's easy to use numpy slicing to pass the segment directly. For read_signal the offset and n_samples parameter get passed right down into the SoundFile call and so can potentially speed things up, i.e. avoid reading a complete signal and then truncating.

clarity/utils/file_io.py
jonbarker68: Yes,  I think I will introduce a separate resample - that takes two sample rates , i.e. just a change to the scipy.resample semantics  -- but then call this from within read_signal().  I'd like to keep it embedded because it's such a common thing to do ... many of our algorithms are designed for specific sample rates. Also the librosa library -- very popular in the community -- has a similar semantic, so it won't seem so odd to people.

PR #296, opened 2023-04-19
Cad1 fix

groadabike:
Hotfix for v0.3.0. Adds `level1` correction to deal with very quiet reference signals. 

Other additions:

- Adds function `compute_rms` in `clarity.utils`
- Adds test function for `compute_rms`
- Add `level1` parameter in `compute_haaqi` function
- Add level1 value when calling `compute_haaqi` in `cad1.task1.baseline.evaluate.py`

### DIFF
--- a/recipes/cad1/task1/baseline/evaluate.py
+++ b/recipes/cad1/task1/baseline/evaluate.py
@@ -1,5 +1,4 @@
 """Evaluate the enhanced signals using the HAAQI metric."""
-# pylint: disable=import-error
 from __future__ import annotations
 
 import csv
--- a/recipes/cad1/task1/baseline/evaluate.py
+++ b/recipes/cad1/task1/baseline/evaluate.py
@@ -1,4 +1,5 @@
 """Evaluate the enhanced signals using the HAAQI metric."""
+# pylint: disable=import-error
 from __future__ import annotations
 
 import csv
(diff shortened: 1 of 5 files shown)


groadabike:
This hotfix will be pushed into `main` in a different way to deal with the different updates.

recipes/cad1/task1/baseline/evaluate.py
```diff
@@ -195,13 +196,15 @@ def _evaluate_song_listener(
...
             config.nalr.fs,
+            65 - 10 * np.log10(compute_rms(left_reference_signal)),
         )
         per_instrument_score[f"right_{instrument}"] = compute_haaqi(
             right_enhanced_signal,
             right_reference_signal,
             np.array(listener_audiograms["audiogram_levels_r"]),
             np.array(listener_audiograms["audiogram_cfs"]),
             config.nalr.fs,
+            65 - 10 * np.log10(compute_rms(right_reference_signal)),
```
jonbarker68, line 207: Shouldn't this be
```
65 - 20 * np.log10(compute_rms(right_reference_signal))
```

i.e. with a factor of 20 rather than 10.
 
The rms value is an amplitude measurement rather than a power measurement.

https://dspillustrations.com/pages/posts/misc/decibel-conversion-factor-10-or-factor-20.html

jonbarker68 (changes requested):
This looks good except that the dB level calculation looks wrong. See comment.

jonbarker68 (approved):
Thanks. Looks good to me.

Issue #301, opened 2023-04-20
introducing dataclass to represent listener


jonbarker68:
# Introducing a Listener dataclass

A listener data class which stores an id and an audiogram for the left and right ear will allow us to simplify recipe code and reduce duplication.

- The class can validate that the left and right ear have audiogram measured at the same frequencies
- It can handle the reading of listener data from the standard clarity metadata files
- It will serve as a useful way to pass parameters to functions that expect left and right ear measurements.

PR #302, opened 2023-04-20
301-introducing-dataclass-to-represent-listener

jonbarker68:
# Introducing a Listener dataclass

A listener data class which stores an id and an audiogram for the left and right ear will allow us to simplify recipe code and reduce duplication.

- The class can validate that the left and right ear have audiogram measured at the same frequencies
- It can handle the reading of listener data from the standard clarity metadata files
- It will serve as a useful way to pass parameters to functions that expect left and right ear measurements.

This PR will close issue #301


clarity/utils/audiogram.py
```diff
@@ -173,3 +175,73 @@ def resample(self, new_frequencies: ndarray) -> Audiogram:
...
+                frequencies=listener_dict["audiogram_cfs"],
+            ),
+            audiogram_right=Audiogram(
+                levels=listener_dict["audiogram_levels_r"],
+                frequencies=listener_dict["audiogram_cfs"],
+            ),
+        )
+
+    @staticmethod
+    def read_listener_dict(filename: Path) -> dict[str, Listener]:
```
groadabike, line 228: An a bit pedantic comment but, maybe called as `read_listener_json` or `read_listener_file` as `read_listener_dict` may be misleading as `dict` is a data structure not a file format.

jonbarker68: Thanks.  I will give this some more thought after integrating into the recipes.  Perhaps read_listener_file is most consistent. The problem here is that in clarity nearly all the metadata has been distributed as a list of dicts -- which is really standard, and reading the files returns a list as expected... but the listeners were for some reason saved in a dictionary structure. So I wanted to highlight that the function returned a dictionary structure, ie. 

```
listeners = read_listener_dict("listeners.json")
```

It might even be best then if this function just flattened the dictionary into a list of values.

jonbarker68: Renamed as "load_listener_dict" which is perhaps more intuitive.

groadabike (approved):
Thank you for dealing with this.

Issue #309, opened 2023-04-26
Alternative resampling methods


jonbarker68:
Expand the resample method in signal_processing.py to offer some alternatives.

Provide

- soxr hiqh quality
- scipy.resample 
- scipy.resample_polyphase

soxr will be the default.

PR #310, opened 2023-04-26
Code and tests for extended resample method

jonbarker68:
## Improved resampling

Improved version of the `signal_processing.resample` method that now supports multiple approaches.

**Note**: the PR will also change the default behaviour of the `file_io.read_signal` function so that resampling is performed with the `soxr` module with the high-quality setting. This matches the behaviour of `librosa.load`

Closes issue #309


groadabike (approved):
Thank you for doing this.
It looks all good to me

PR #318, opened 2023-05-11
316 conflicts in torch devices in torchloudnorm amd convtasnet

### LINKED ISSUE #316: Conflicts in torch devices in torchloudnorm amd ConvTasNet
`torchloudnorm` doesn't have a parameter to set the device, instaed, is selected by default by using `cuda` if found or `cpu`. This can result in device conflicts when a GPU is available but the tensors are on `cpu`.

 `ConvTasNet` have similar problems when `cpu` is selected

groadabike:
closes #316 

- Add device parameter in torchloudnorm
- Use the device in add_and_overlap in ConvTasNet
- Selects the resampling method according to the first number of the torchaudio version.
    - V0.xx options are ["sinc_interpolation", "kaiser_window"] 
    - V2.x.x options are [sinc_interp_hann, sinc_interp_kaiser]

### DIFF
--- a/recipes/cec1/e009_sheffield/test.py
+++ b/recipes/cec1/e009_sheffield/test.py
@@ -25,15 +25,21 @@ def run(cfg: DictConfig) -> None:
 
     down_sample = up_sample = None
     if cfg.downsample_factor != 1:
+        if torchaudio.__version__[0] == "0":
+            # For versions v0.x.x
+            resampling_method = "sinc_interpolation"
+        else:
+            # for versions v2.x.x
+            resampling_method = "sinc_interp_hann"
         down_sample = torchaudio.transforms.Resample(
             orig_freq=cfg.sample_rate,
             new_freq=cfg.sample_rate // cfg.downsample_factor,
-            resampling_method="sinc_interp_hann",
+            resampling_method=resampling_method,
         )
         up_sample = torchaudio.transforms.Resample(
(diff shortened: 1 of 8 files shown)


groadabike:
@jonbarker68 
I can see that the reduction in the codecod % is due to:
1. The if-else statement to validate the torchaudio version to set the default `resampling_method`
2. The if-else for the torch device selection

recipes/cec1/e009_sheffield/test.py
```diff
@@ -25,15 +25,21 @@ def run(cfg: DictConfig) -> None:
 
     down_sample = up_sample = None
     if cfg.downsample_factor != 1:
+        if torchaudio.__version__[0] == "0":
```
jonbarker68, line 28: I'd rather we just drop the support for the v0 torch audio. By the time we release v0.4 torch 2.0 and the new torchaudio will have been available a long time. The dependencies already have torch>=2

jonbarker68 (changes requested):
Thanks.  All OK but can we drop support for v0.x.x. of torchaudio and remove the logic that is checking torchaudio.__version__

groadabike:
> Thanks. All OK but can we drop support for v0.x.x. of torchaudio and remove the logic that is checking torchaudio.**version**

Yeah, no problem.
I think you are right. I did it because all the tests were failing on my laptop because my local Python env is 0.3.2.
But, to make changes in the `main branch` I'll need to create an updated environment

PR #346, opened 2023-09-17
Fixed smear3 to work with newer numpy versions

jonbarker68:
This should fix the problem that was causing some tests to fail with the newer numpy versions.


jonbarker68:
@groadabike this fixes the numpy 1.25 and 1.26 issues

groadabike:
That's great, thank you for this.

Sunsunny11 (approved):
Thanks.

PR #364, opened 2024-02-20
Refactoring HAAQI to save precomputations

### LINKED ISSUE #363: Add precomputations to HAAQI
There are two banches working with HAAQI to improve the metric. However, after several analysis, I realised that many of the slowest parts can be improved by precomputing values. These values correspond computations that are not related to the signals. Additionally, as HAAQI works at 24000 Hz, many of the computations that depend on the sample rate can also be precomputed:

1. Filter coefficients in `eb.env_compress_basilar_membrane`: 
     Depends on a constant (flp = 800) and sample rate

2. Filter coefficients in `eb.middle_ear` 
     Depends on sample rate 

3. Group delay vector in `eb.group_delay_compensate`
     These are constant

4. `Sin` and `Cos` in `eb.gammatone_bandwidth_demodulation`
     This is a recursive function (every 300 samples) that depends on the length of the audio signal (i.e., 24000 samples per second of audio) and the sample rate. However, considering values to the 10th decimal place, the function is periodic. 

image

image

groadabike:
closes #363 

This is the first draft of the refactoring of HAAQI to precompute and reuse some parts of the code.
This logic can the be applied to HASPI and HASQI

Need to run a more extensive sanity test to ensure the consistency of the output compared with the old implementation 

TODO:
- [x] Run the metric several times to measure consistency with old implementation
- [x] Create unit and regression tests


groadabike:
@jonbarker68, this is a draft of this refactoring.
if you have a minute, can you take a look to see if makes sense to you?

groadabike:
The output of this refactoring is not exactly the same as previous implementation due to the different order in what random noise is generated.

groadabike:
After a lot of work trying to improve HAAQI computation times, this approach didn't help.
This needs rethinking
Closing the PR.

PR #376, opened 2024-05-16
374 multiband dynamic range compressor

### LINKED ISSUE #374: Multiband Dynamic Range Compressor

groadabike:
Implementation of the Multiband Compressor to use in CAD2

Closes #374


groadabike:
@jonbarker68 
I included the knee width parameter in the comprressor by replacing equation 3 for equation 4 from the paper 
https://www.eecs.qmul.ac.uk/~josh/documents/2012/GiannoulisMassbergReiss-dynamicrangecompression-JAES2012.pdf
image

image

Can you plase take a look and see if there is something wrong?

Cheers

clarity/enhancer/multiband_compressor/crossover.py
```diff
@@ -0,0 +1,210 @@
...
+    def __call__(self, signal: np.ndarray) -> np.ndarray:
+        """Apply the crossover filter to the signal.
+
+        Args:
+            signal (np.ndarray): The input signal.
+
+        Returns:
+            np.ndarray: The filtered signal.
+        """
+        # Running only with 5 crossover filters
```
jonbarker68, line 150: Why is this specific to 5 filters?

clarity/enhancer/multiband_compressor/crossover.py
```diff
@@ -0,0 +1,210 @@
...
+            freq_crossover (list | ndarray | int): The crossover frequency (Hz).
+            N (int): The order of the filter.
+            sample_rate (float): The sample rate of the signal (Hz).
+        """
+        if isinstance(freq_crossover, int):
+            freq_crossover = [freq_crossover]
+
+        if N != 4:
+            raise ValueError(f"The order of the filter must be 4. {N} was provided.")
+
```
jonbarker68, line 53: Why do we need this? Code looks general for any even order.  Otherwise, if it is only correct for 4th order wouldn't it be clearer to remove the N parameter and set it as a constant?

clarity/enhancer/multiband_compressor/crossover.py
```diff
@@ -0,0 +1,210 @@
...
+        self.bstore = np.zeros((self.order + 1, self.xover_freqs.shape[0], 2))
+        self.astore = np.zeros((self.order + 1, self.xover_freqs.shape[0], 2))
+        self.bstore_phi = np.zeros((2 * self.order + 1, self.xover_freqs.shape[0]))
+        self.astore_phi = np.zeros((2 * self.order + 1, self.xover_freqs.shape[0]))
+        self.compute_coefficients()
+
+    def compute_coefficients(self) -> None:
+        """Compute the filter coefficients.
+        These are independednt of the signal and can be computed once.
+        """
```
jonbarker68, line 69: typo

clarity/enhancer/multiband_compressor/crossover.py
```diff
@@ -0,0 +1,210 @@
...
+        self.sample_rate = sample_rate
+
+        # Initialize the filter coefficients
+        self.bstore = np.zeros((self.order + 1, self.xover_freqs.shape[0], 2))
+        self.astore = np.zeros((self.order + 1, self.xover_freqs.shape[0], 2))
+        self.bstore_phi = np.zeros((2 * self.order + 1, self.xover_freqs.shape[0]))
+        self.astore_phi = np.zeros((2 * self.order + 1, self.xover_freqs.shape[0]))
+        self.compute_coefficients()
+
+    def compute_coefficients(self) -> None:
```
jonbarker68, line 66: Could be a function taking xover_freqs, order and sample rate and returning the bstore_phi and astore_phi values. This would also mean you could remove lines 62 and 63 from the constructor. i.e. their values are set directly from the return of the compute_coefficients function

clarity/enhancer/multiband_compressor/multiband_compressor.py
```diff
@@ -0,0 +1,259 @@
...
+                (bands, channel, samples)
+        """
+
+        bands_signal = self.crossover(signal[0])[:, np.newaxis, :]
+        if signal.shape[0] > 1:
+            for sig in signal[1:]:
+                bands_signal = np.hstack(
+                    (bands_signal, self.crossover(sig)[:, np.newaxis, :])
+                )
+
```
jonbarker68, line 240: I don't think this loop should be necessary. Your crossover filter is based on lfilter and so should be able to apply the crossover to all the channels in a single call.   i.e. bands_signal = self.crossover(signal)[:, np.newaxis, :].  You might need to make sure that if the signal is single channel then an axis added. And you will need to modify crossover so that it filters along the time axes.

jonbarker68 (dismissed):

jonbarker68:
Hi Gerardo, this looks great. I've added a couple of suggestions in the review that will simplify at little.

clarity/enhancer/multiband_compressor/crossover.py
groadabike: This comment was old, that was true before I generalised the number of xovers.
Comment deleted.

clarity/enhancer/multiband_compressor/crossover.py
groadabike: The code only works for order 4,  I tried testing other orders and got bad results.
Changing to constant

jonbarker68 (approved):
Looks great! Let's merge this.

PR #378, opened 2024-05-17
Causal and non-causal vocal-background separation using ConvTasNet

groadabike:
This is the recipe to run the lyrics/accompaniment separtion models


groadabike:
@jonbarker68 any idea what can be the best way to test this?

groadabike:
Add Model to CAD2 branch

PR #413, opened 2024-09-11
Jpb/support for python 312

jonbarker68:
Rolling Python support forward one version:
- Adding support from python3.12 and dropping for 3.8. i.e. tested against 3.9, 3.10, 3.11, 3.12
- Upgrading to numpy 2.x, pandas 2.x, and newer scipy that introduces scipy.signals.windows
- Relaxed some regression tests that were failing on the mac due to very small numerical differences


groadabike (approved):
It looks great.
We could fix the Codacy error/warnings but there are a couple in that I don't understand.

recipes/cec2/baseline/evaluate.py
groadabike, line 1: Do we want to add the docstring here before merge?

groadabike:
All good. Merged.

PR #419, opened 2024-10-15
Better Intelligibility metric for CAD2 Task1

groadabike:
In CAD2 task1, the original intelligibility metric was using whisper for transcription and jiwer for computing correctness.
However, jiwer was used without any normalisation resulting in lower score.
For example, capitalised vs non-capitalised words, punctuations not excluded.

The changes are:

Replace jiwer by alt-eval which performs several normalisation before calling jiwer.
Add verbose option True or False for MSGB HL model to reduce unnecessary prints. Warnings will still be printed regardless the verbose option.

### DIFF
--- a/clarity/evaluator/msbg/cochlea.py
+++ b/clarity/evaluator/msbg/cochlea.py
@@ -15,6 +15,7 @@ from clarity.evaluator.msbg.msbg_utils import read_gtf_file
 from clarity.evaluator.msbg.smearing import Smearer
 from clarity.utils.audiogram import Audiogram
 
+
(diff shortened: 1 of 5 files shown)


sgraetzer:
> In CAD2 task1, the original intelligibility metric was using whisper for transcription and jiwer for computing correctness. However, jiwer was used without any normalisation resulting in lower score. For example, capitalised vs non-capitalised words, punctuations not excluded.
> 
> The changes are:
> 
> Replace jiwer by alt-eval which performs several normalisation before calling jiwer. Add verbose option True or False for MSGB HL model to reduce unnecessary prints. Warnings will still be printed regardless the verbose option.

Excellent.

PR #458, opened 2026-07-06
Clip2 recipe

groadabike:
Baseline recipe for CLIP2.
A DNN model that uses whisper-large encoder for signal embedding and a regression head to predict intelligibility
Is has a dataloader, model, train and test scripts


jonbarker68 (commented):
Great! Thanks

jonbarker68 (approved):

